Issue #1, opened 2020-02-29
[firmware-upgrader] Set up travis build and qa checks


nemesisdesign:
Set up a travis build in line with the other openwisp modules (see django-freeradius, openwisp-ipam, for recent good examples).

The build should test the latest versions of django (2.2 and 3.0) and perform all the qa-checks done in the other repos.

PR #6, opened 2020-03-02
[firmware-upgrader] Update depdendencies

### LINKED ISSUE #2: [firmware-upgrader] Update depdendencies
Upgrade all the dependencies.
Drop support for python 2 code.

nepython:
Closes #2


nepython:
Don't merge it yet, a bit of relevant lines needs to be changed.

.travis.yml
```diff
@@ -12,8 +12,13 @@ addons:
       - libspatialite-dev
 
 python:
-  - "3.4"
-  - "2.7"
+  - "3.5"
```
IAmAMissingColon: I think you should remove `Python 3.5` :smile:

nepython: I was going to ask that whether it's needed, because if not I can enhance it a bit further, eg introducing f-strings :smile:

atb00ker: Yes, we are currently moving to python 3.6 and + ! :smile:

.travis.yml
```diff
@@ -27,11 +32,9 @@ before_install:
 
 install:
   - python setup.py -q develop
-  - if [[ $TRAVIS_PYTHON_VERSION == "2.7" ]]; then pip install git+git://github.com/tinio/pysqlite.git@extension-enabled#egg=pysqlite; fi
 
 script:
   - coverage run --source=openwisp_firmware_upgrader runtests.py
```
atb00ker: Please add the standard `openwisp-utils` qa checks as well! :smile:

atb00ker: (check the other repositories's travis.yml for it), example : 
```
openwisp-utils-qa-checks \
    --migration-path \
      "./openwisp_controller/config/migrations \
      ./openwisp_controller/connection/migrations \
      ./openwisp_controller/geo/migrations \
      ./openwisp_controller/pki/migrations" \
    --migrations-to-ignore "12 0 0 4" \
    --migration-module "config connection geo pki";
```
from controller: https://github.com/openwisp/openwisp-controller/blob/master/.travis.yml

IAmAMissingColon: Hello ! @atb00ker I saw separate  [issue](https://github.com/openwisp/openwisp-firmware-upgrader/issues/1) on it :smile: can I work on it after finishing my current one ?

nemesisdesign: Let's do the improvement of the build (qa-checks and so on) in the other issue

.travis.yml
```diff
@@ -27,11 +32,9 @@ before_install:
...
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.4 ]]; then ./tests/manage.py makemigrations --dry-run | grep "No changes detected"; fi
```
nemesisdesign: don't remove the entire line, remove only the if

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -1,4 +1,4 @@
-import mock
+from unnittest.mock import patch
```
nemesisdesign: `from unnittest import mock` and avoid changing all the other lines because it's not necessary nor brings any value to do so (actually mock.patch is slightly more explicit than just patch).

tests/openwisp2/celery.py
```diff
@@ -1,11 +1,11 @@
...
 
+from django.conf import settings
+
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'openwisp2.settings')
 
 app = Celery('openwisp2')
-app.config_from_object('django.conf:settings', namespace='CELERY')
-app.autodiscover_tasks()
+app.config_from_object('django.conf:settings', namespace = 'CELERY')
+app.autodiscover_tasks(settings.INSTALLED_APPS)
```
nemesisdesign: revert the changes on these two lines please, see also:
https://github.com/celery/celery/issues/3341#issuecomment-235408048

nepython:
@nemesisdesign, I have implemented all suggested changes yet, build is failing due to unsuccessful migrations, might be fixed by PR #7. The exact cause of migration points to namespace in celery, I checked with their github page and the code looks perfectly fine to me.

**Traceback** 
```
  File "/home/travis/build/openwisp/openwisp-firmware-upgrader/tests/openwisp2/celery.py", line 8, in <module>
    app.config_from_object('django.conf:settings', namespace='CELERY')
TypeError: config_from_object() got an unexpected keyword argument 'namespace'
```
 Besides that can you please review the code.
:)

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -82,7 +82,7 @@ def _reflash(self, path):
         """
         def upgrade(conn, path, timeout):
             conn.connect()
-            conn.exec_command('sysupgrade -v -c {0}'.format(path),
+            conn.exec_command(f'sysupgrade -v -c {path}',
```
nemesisdesign: do not use f-strings because we still want to support python 3.6 (f-strings are available from python 3.7 onwards)

requirements.txt
```diff
@@ -0,0 +1,4 @@
+djangorestframework>=3.11.0,<3.12
+openwisp-utils>=0.4.4,<0.5.0
+pyasn1>=0.4.8,<0.5.0
+django-celery>=3.3.1,<3.4.0
```
nemesisdesign: why are you adding this file here? It's redundant

setup.py
```diff
@@ -49,7 +49,10 @@ def get_install_requires():
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        'openwisp-controller>=0.3.1,<0.4.1'
+        'celery>=4.4.0,<5.0',
+        'openwisp-controller>=0.5.1,<0.6.1',
+        'openwisp-utils>=0.4.4,<0.5.0',
+        'djangorestframework>=3.11.0,<3.12'
```
nemesisdesign: why are you adding all these dependencies here? openwisp-controller already depends on those so it should not be needed to re-add those here

.travis.yml
```diff
@@ -22,16 +28,16 @@ branches:
 before_install:
   - pip install -U pip wheel setuptools
   - pip install --no-cache-dir -U -r requirements-test.txt
+  - pip install --no-cache-dir -U -r requirements.txt
```
nemesisdesign: remove this, not needed. Dependencies will be installed when doing `python setup.py -q develop`

setup.py
```diff
@@ -49,7 +49,10 @@ def get_install_requires():
     include_package_data=True,
     zip_safe=False,
     install_requires=[
-        'openwisp-controller>=0.3.1,<0.4.1'
+        'celery>=4.4.0,<5.0',
+        'openwisp-controller>=0.5.1,<0.6.1',
```
nemesisdesign: should be `openwisp-controller>=0.5.0,<0.6.0`

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -1,6 +1,6 @@
-import mock
 from django.core.exceptions import ValidationError
 from django.test import TestCase, TransactionTestCase
+from unnittest import mock
```
nemesisdesign: `unnittest` > `unittest` (2 N letters, should be 1)

tests/openwisp2/settings.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import absolute_import, unicode_literals
```
nemesisdesign: this should not be needed, why are you adding it?

requirements-test.txt
```diff
@@ -1,8 +1,8 @@
+celery>=4.4.0,<5.0
```
nemesisdesign: remove this, it's already defined in openwisp-controller which is a dependency of this package

openwisp_firmware_upgrader/upgraders/openwrt.py
nepython: I think they are supported from 3.6 onwards.
 **Source:**[Python docs: What's new in 3.6](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498)
I tested it locally on my python version 3.6.9 and it does work. Please correct me if I am wrong.

requirements.txt
nepython: @nemesisdesign I am currently getting the following in my build logs after removing all unrequired lines. Also, I was previously getting the same traceback which led me to add them.
```
error: The 'djangorestframework' distribution was not found and is required by djangorestframework-gis
The command "python setup.py -q develop" failed and exited with 1 during 
```
& rest of the modules which I had added came up with similar error once this is resolved.
So can you please suggest edits to resolve this.

.travis.yml
```diff
@@ -27,11 +33,10 @@ before_install:
 
 install:
   - python setup.py -q develop
```
nemesisdesign: try to use the following: `pip install -U -e .`

openwisp_firmware_upgrader/upgraders/openwrt.py
nepython: @nemesisdesign, what about this? 😅

nemesisdesign: You're right, from python 3.6 onwards :+1:

.travis.yml
nepython: I didn't knew about it earlier. Thanks a lot, finally learnt a way to install dependencies.:smile: 
I have updated the code as per your suggestions.

nepython:
@nemesisdesign, two major reasons for test failures are:
1)`FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/openwisp/openwisp-firmware-upgrader/tests/openwisp2/media/fake-img.bin'`
2)`redis.exceptions.ConnectionError: Error 111 connecting to localhost:6379. Connection refused.`
In the first case I am unaware whether the file mention has to be removed or the file is yet to be added while in the second case full traceback points to the name,key for the connection already existing, leading to it's likely refusal.
https://github.com/openwisp/openwisp-firmware-upgrader/blob/61af6f95e6437dc8e70d4eca60772f6a33c0996c/openwisp_firmware_upgrader/tests/test_admin.py#L17-L22
As for creating any new login the same function is being used, thus same values. This is what I think is most probably the reason:smile:

nepython:
@nemesisdesign, I looked up in the Django docs and found everything perfect with what is already there in current code(Though the docs officially showed it for MemCached). So, I didn't quite get what is being missed though I referred multiple sources and tried many changes locally. Finally on commiting this:
  `'KEY_PREFIX': 'example'` (Though I don't think this was required as many examples worked without it too.)
Now the error coming up in the build after this line's addition is 
`# No valid IP addresses to initiate connections found`
While I have implemented the first change suggested by you. I would be very grateful if you can suggest edits as I too am trying out memory caching for the first time :smile:

PR #20, opened 2020-04-07
[upload] Store images in private location #13

### LINKED ISSUE #13: [upload] Store images in private location
Right now firmware images are stored in the media folder which is publicly accessible via unauthenticated HTTP (provided one knows the exact URL), we have to find a way to achieve the following:

- firmware images have to be stored in a directory which is not accessible via unauthenticated HTTP requests
- in order to download a file the user must be either superuser or must be a staff user with access to the organization
- if possible, we should put all images of a build in a dedicated directory, this also avoids name collisions, it seem that now images are all saved in one directory, which will most probably cause name collisions

This open source package should help us in achieving this more quickly: https://github.com/edoburu/django-private-storage

PabloCastellano:
Use django-private-storage to make media uploaded files available to staff users only
Only superusers or staff members that belong to the organization can access and download
firmware images.
Firmware images are uploaded to firmware/<build_uuid>/<filename>
Added a new setting OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE which
allows defining a maximum size for firmware images. By default it is None = unlimited.

Closes #13


openwisp_firmware_upgrader/base/models.py
```diff
@@ -127,7 +128,7 @@ def _find_firmwareless_devices(self, boards=None):
 class AbstractFirmwareImage(TimeStampedEditableModel):
     build = models.ForeignKey(get_model_name('firmware_upgrader', 'Build'),
                               on_delete=models.CASCADE)
-    file = models.FileField()
+    file = PrivateFileField("File")
```
nemesisdesign, line 131: use single quotes for consistency

setup.py
```diff
@@ -34,8 +34,9 @@
     include_package_data=True,
     zip_safe=False,
     install_requires=[
+        'django-private-storage==2.2.2',
```
nemesisdesign, line 37: use a dependency range, eg `~=2.2`(will install latest version in 2.2.x), I'd move it after openwisp-controller, is there a reason you put it first?

PabloCastellano: Yeah, to keep it alphabetically sorted. Is there any preference?

PabloCastellano:
Not sure why travis is failing, since I've added the migration file

setup.py
nemesisdesign: Main openwisp dependencies first and then the rest

openwisp_firmware_upgrader/base/models.py
```diff
@@ -7,14 +7,13 @@
...
 from django.utils.module_loading import import_string
 from django.utils.translation import ugettext_lazy as _
+from private_storage.fields import PrivateFileField
 from swapper import get_model_name, load_model
 
 from openwisp_controller.config.models import Device
 from openwisp_users.mixins import OrgMixin
 from openwisp_utils.base import TimeStampedEditableModel
 
-from private_storage.fields import PrivateFileField
```
nemesisdesign: for these kind of changes, use --amend and then push --force

PabloCastellano:
> Good progress, does the download of the images by authenticated users already work?

Yes, it works. If the user is not authenticated he will get a plain 403 Forbidden page. Also if the file does not exist, this way it avoids file enumeration by anonymous users.

image

> What happens if you upload an image and then click on the link in the admin, does it let you download the file?

Yes, just like a normal media file.

The only big thing to mention in this change is that you must move the files in your `MEDIA_ROOT` to the new `PRIVATE_STORAGE_ROOT` path, otherwise the link will not work because the file does not exist in that path. We should somehow mention it in the next release notes.

openwisp_firmware_upgrader/settings.py
```diff
@@ -6,3 +6,6 @@
 UPGRADERS_MAP = getattr(settings, 'OPENWISP_FIRMWARE_UPGRADERS_MAP', {
     conn_settings.DEFAULT_UPDATE_STRATEGIES[0][0]: 'openwisp_firmware_upgrader.upgraders.openwrt.OpenWrt'
 })
+
+PRIVATE_STORAGE_CLASS = 'openwisp_firwmare_upgrader.private_storage.storage.OpenWISPPrivateFileSystemStorage'
+OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE = getattr(settings, 'OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE', None)
```
nemesisdesign, line 11: can you default this to 30 MB please?

openwisp_firmware_upgrader/private_storage/views.py
```diff
@@ -0,0 +1,22 @@
...
+    model_file_field = 'file'
+
+    slug_field = 'file'
+    slug_url_kwarg = 'file'
+
+    def can_access_file(self, private_file):
+        user = private_file.request.user
+        if not user.is_authenticated:
+            return False
+        image = FirmwareImage.objects.get(file=private_file.relative_name)
```
nemesisdesign, line 16: if you try to access use the `get_object` method of `PrivateStorageDetailView`, what comes out?

openwisp_firmware_upgrader/private_storage/storage.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import reverse_lazy
+from private_storage.storage.files import PrivateFileSystemStorage
+
+
+class OpenWISPPrivateFileSystemStorage(PrivateFileSystemStorage):
```
nemesisdesign, line 5: Let's name it `FirmwarePrivateFileSystemStorage`

openwisp_firmware_upgrader/base/models.py
```diff
@@ -124,10 +125,17 @@ def _find_firmwareless_devices(self, boards=None):
                                      model__in=boards)
 
 
+def get_build_directory(instance, filename):
+    build_id = str(instance.build.id)
```
nemesisdesign, line 129: use `pk` because developers extending this module may replace the uuid with a different kind of primary key (it happened once in django-x509).

openwisp_firmware_upgrader/settings.py
nemesisdesign: another thing, for consistency it should be:

`MAX_FILE_SIZE = getattr(settings, 'OPENWISP_FIRMWARE_UPGRADER_MAX_FILE_SIZE', None)`

openwisp_firmware_upgrader/settings.py
```diff
@@ -6,3 +6,6 @@
 UPGRADERS_MAP = getattr(settings, 'OPENWISP_FIRMWARE_UPGRADERS_MAP', {
     conn_settings.DEFAULT_UPDATE_STRATEGIES[0][0]: 'openwisp_firmware_upgrader.upgraders.openwrt.OpenWrt'
 })
+
+PRIVATE_STORAGE_CLASS = 'openwisp_firwmare_upgrader.private_storage.storage.OpenWISPPrivateFileSystemStorage'
```
nemesisdesign, line 10: let's make this configurable as well, since we're here, it doesn't cost much.

tests/openwisp2/settings.py
```diff
@@ -96,6 +97,8 @@
 STATIC_URL = '/static/'
 MEDIA_URL = '/media/'
 MEDIA_ROOT = '{0}/media/'.format(BASE_DIR)
+PRIVATE_STORAGE_ROOT = '/{0}/firmware/'.format(BASE_DIR)
+PRIVATE_STORAGE_CLASS = 'openwisp_firmware_upgrader.private_storage.storage.OpenWISPPrivateFileSystemStorage'
```
nemesisdesign, line 101: why is this setting here? Is it needed for django-private-storage?

nemesisdesign: I'd set those two settings as defaults in https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/apps.py#L12, allowing them to be overridden.
I'll do it later if you don't get to do it now.

openwisp_firmware_upgrader/private_storage/views.py
PabloCastellano: Used self.object instead :+1:

PR #21, opened 2020-04-10
Add default permissions for operators and administrators #15

### LINKED ISSUE #15: [fw-upgrade] Add default permissions
Operator:

- view builds and related images
- view mass upgrade operations and related device operations
- view device firmware objects

Administrator:

- all permission on all objects

Example of how we did it in another module: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/migrations/0004_default_permissions.py

PabloCastellano:
Operators can:

* view builds and related images
* view mass upgrade operations and related device operations
* view device firmware objects

Administrators can:
* all permission on all objects

Closes #15

### DIFF
--- a/openwisp_firmware_upgrader/migrations/0002_default_permissions.py
+++ b/openwisp_firmware_upgrader/migrations/0002_default_permissions.py
@@ -1,0 +1,60 @@
+from django.db import migrations
+from django.contrib.auth.models import Permission
+
+from . import create_default_permissions
+
+
+
+def assign_permissions_to_groups(apps, schema_editor):
+    app_label = 'firmware_upgrader'
+    create_default_permissions(apps, schema_editor)
+    Group = apps.get_model('openwisp_users', 'Group')
+
+    try:
+        admin = Group.objects.get(name='Administrator')
+        operator = Group.objects.get(name='Operator')
+    # consider failures custom cases
+    # that do not have to be dealt with
+    except Group.DoesNotExist:
+        return
+
+    operators_read_only_admins_manage = ['build', 'devicefirmware', 'firmwareimage', 'batchupgradeoperation', 'upgradeoperation']
+    admins_can_manage = ['category']
+    manage_operations = ['add', 'change', 'delete']
+
+    for action in manage_operations:
+        for model_name in admins_can_manage:
+            permission = Permission.objects.get(
+                codename='{}_{}'.format(action, model_name),
+                content_type__app_label=app_label
+            )
+            admin.permissions.add(permission.pk)
+    for model_name in operators_read_only_admins_manage:
+        try:
+            permission = Permission.objects.get(
+                    codename='view_{}'.format(model_name),
+                    content_type__app_label=app_label
+            )
+            operator.permissions.add(permission.pk)
+        except Permission.DoesNotExist:
+            pass
+        for action in manage_operations:
+            permission_ad = Permission.objects.get(
+                    codename='{}_{}'.format(action, model_name),
+                    content_type__app_label=app_label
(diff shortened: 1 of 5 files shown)


openwisp_firmware_upgrader/migrations/0003_default_permissions.py
```diff
@@ -0,0 +1,52 @@
+from django.db import migrations
+from django.contrib.auth.models import Permission
+
+from . import create_default_permissions
+
+
+
+def assign_permissions_to_groups(apps, schema_editor):
```
nemesisdesign, line 8: a similar migration should be created for the sample app as well or we'll have two different test environments.

Since this operation will have to be done not only by the sample_app but potentially by any app that will extend this module, it's good practice to make this function reusable by passing `app_label` as a parameter.

The actual function passed to `RunPython` will call the reusable function.

openwisp_firmware_upgrader/migrations/0003_default_permissions.py
```diff
@@ -0,0 +1,52 @@
...
+        return
+
+    operators_read_only_admins_manage = ['build', 'devicefirmware', 'firmwareimage', 'batchupgradeoperation', 'upgradeoperation']
+    admins_can_manage = ['category']
+    manage_operations = ['add', 'change', 'delete']
+
+    for action in manage_operations:
+        for model_name in admins_can_manage:
+            permission = Permission.objects.get(
+                codename='{}_{}'.format(action, model_name)
```
nemesisdesign, line 27: this is missing:

```
content_type__app_label=app_label
```

Please add it here and to all the other `Permission.objects.get` calls

openwisp_firmware_upgrader/migrations/0003_default_permissions.py
```diff
@@ -0,0 +1,52 @@
...
+            pass
+        for action in manage_operations:
+            permission_ad = Permission.objects.get(codename='{}_{}'.format(action, model_name))
+            admin.permissions.add(permission_ad.pk)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('firmware_upgrader', '0001_initial'),
```
nemesisdesign, line 44: this is migration number 0003 and is pointing to 0001, that's because 0002 is the private storage field.
Is this ok? Or should the order be changed depending on which one will be merged first?

PabloCastellano:
Fixed:

* Migration 0002 instead of 0003
* Added `content_type__app_label=app_label`
* Added permissions testcase

openwisp_firmware_upgrader/migrations/0003_default_permissions.py
PabloCastellano: We should probably fix it at https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/migrations/0004_default_permissions.py too

openwisp_firmware_upgrader/migrations/0002_default_permissions.py
```diff
@@ -0,0 +1,60 @@
+from django.db import migrations
+from django.contrib.auth.models import Permission
+
+from . import create_default_permissions
+
+
+
+def assign_permissions_to_groups(apps, schema_editor):
```
nemesisdesign, line 8: this function can be shared between the two migrations, I will show you

nemesisdesign: Please see https://github.com/openwisp/openwisp-firmware-upgrader/commit/079cb2644da56cf1191904becde2d5f5341ebd81

PR #31, opened 2020-04-15
[ci] Run sample app tests in same build to reduce build time

nemesisdesign:
I'm testing this change to save build time since creating a new container and installing all the packages requires time and resources on the side of travis. We'll just have to double check:

1. sample app tests are run as expected
2. coverage is not affected and works properly

@atb00ker @nepython @PabloCastellano @NoumbissiValere @TheOneAboveAllTitan please check this out before it will affect part of your work.

If this approach work we should adopt it on all the modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,8 @@ before_install:
 
 install:
   - pip install -e .
+  # temporary: remove when openwisp-utils 0.5.0 is released
+  - pip install -U https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils[qa]
 
 script:
   - SAMPLE_APP=1; coverage run --source=openwisp_firmware_upgrader runtests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,6 @@ python:
 env:
   - DJANGO="django>=2.2,<2.3"
   - DJANGO="django>=3.0,<3.1"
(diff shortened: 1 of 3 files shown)


PabloCastellano:
@nemesisdesign The approach looks good to me and here are some comments:

* Why are you using `--skip-black` only in the sample app?
* exporting SAMPLE_APP=1 and unsetting it is not required. The following works too:

```
coverage run --source=openwisp_firmware_upgrader runtests.py
SAMPLE_APP=1 coverage run --source=openwisp_firmware_upgrader runtests.py
`` 

(you can double check it by adding -v2 at the end of the command)

nemesisdesign:
@PabloCastellano:

- I'm skipping most checks in the second `openwisp-qa-check` call because they are already done in the first call, so the second call can be dedicated to only checking the migrations of the sample app
- I am not sure I understand why unsetting is required, is it because the env of each command is separated from the previous one? Nonetheless, if the unsetting is present, it won't harm and it will be obvious to those who don't know the internals of travis right?

PabloCastellano:
@nemesisdesign Regarding 2) it doesn't hurt but I think it is better to keep it simpler. Up to you in any case :+1: . About running scripts with environment variables here it is very well explained with examples: https://www.baeldung.com/linux/pass-command-line-arguments-bash-script#environment-variables

nemesisdesign:
ok @PabloCastellano I understood what you meant, I removed the usage of export to simplify the solution

PR #33, opened 2020-04-16
[upload] Remove build dir if empty #32

### LINKED ISSUE #32: [fw-upgrade] Remove build dir if empty
Since #20, files will be uploaded in a subdirectory which uses the build PK as its name.

Now when files are deleted, empty build directories remain (visible after running tests as well).

We need to update this method to handle this case: https://github.com/openwisp/openwisp-firmware-upgrader/blob/1e9fa26d75a48ae9274c2708ea9a10134a2bb9d0/openwisp_firmware_upgrader/base/models.py#L200-L206

I'd check whether the directory containing the file to remove is named as the `build.pk` and if yes and if the directory is empty, remove the directory as well.

PabloCastellano:
Fixes #32

### DIFF
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -186,6 +186,7 @@ class AbstractFirmwareImage(TimeStampedEditableModel):
     def delete(self, *args, **kwargs):
         super().delete(*args, **kwargs)
         self._remove_file()
+        self._remove_empty_directory()
 
     def _clean_type(self):
         """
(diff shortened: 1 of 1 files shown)


openwisp_firmware_upgrader/base/models.py
```diff
@@ -205,6 +206,11 @@ def _remove_file(self):
             msg = 'firmware image not found while deleting {0}:\n{1}'
             logger.error(msg.format(self, path))
 
+    def _remove_empty_directory(self):
+        path = os.path.dirname(self.file.path)
+        if not os.listdir(path):
```
nemesisdesign, line 211: could you double check that path is not equal to `settings.MEDIA_ROOT`? Just to be sure we won't accidentally delete that, since we used it previously (I'd add a comment with this info and a TODO: remove this in first release)

PR #37, opened 2020-04-21
[api] Add REST API #26

### LINKED ISSUE #26: [api] Add REST API
Depends on https://github.com/openwisp/openwisp-users/issues/104.

This is an optional feature which should be disabled by default, the setting to use shall be `OPENWISP_FIRMWARE_UPGRADER_API`.

Add django-rest-framework in the `extra_requires['rest']` of setup.py. 

Add the API endpoints to perform the following operations, they shall use the TokenAuthentication defined in openwisp-users and use the DjangoPermission set and users shall be able to access only the objects they're authorized to see based on their organization (superusers can access everything, staff users can access only objects related to organizations they're member of and flagged as is_admin):

- Build:
  - view list of builds (possibility to filter by org)
  - add (should be possible to upload firmware images as well)
  - change (should be possible to add/change/remove firmware images to the build as well)
  - delete
- Category
  - view (possibility to filter by org)
  - add
  - change
  - delete
- Mass upgrade operations
  - view (possibility to filter by org)
  - run mass upgrade

PabloCastellano:
Requires
* https://github.com/openwisp/openwisp-users/pull/109
* https://github.com/openwisp/openwisp-users/pull/105
* https://github.com/openwisp/openwisp-controller/pull/194

Use OPENWISP_FIRMWARE_UPGRADER_API to enable/disable it

Closes #26


openwisp_firmware_upgrader/api/serializers.py
```diff
@@ -0,0 +1,24 @@
...
+        fields = '__all__'
+
+
+class CategorySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Category
+        fields = '__all__'
+
+
+class UpgradeOperationSerializer(serializers.ModelSerializer):
```
nemesisdesign, line 21: shouldn't this be `BatchUpgradeOperation`? The mass upgrade operation should contain the info about the related upgrade operations in its detail view.

openwisp_firmware_upgrader/apps.py
```diff
@@ -8,8 +8,16 @@ class FirmwareUpdaterConfig(AppConfig):
     label = 'firmware_upgrader'
     verbose_name = _('Firmware Management')
 
+    rest_framework_defaults = {
+        'DEFAULT_AUTHENTICATION_CLASSES': [
+            'openwisp_users.api.authentication.BearerAuthentication',
+        ],
+        'DEFAULT_PERMISSION_CLASSES': ['rest_framework.permissions.IsAuthenticated'],
+    }
+
```
nemesisdesign, line 17: this could be something good to move to openwisp-users when the API is active, so we can avoid having to re-define this in every module.

tests/openwisp2/settings.py
```diff
@@ -52,7 +52,9 @@
     'leaflet',
     # rest framework
     'rest_framework',
+    'rest_framework.authtoken',
     'rest_framework_gis',
+    'drf_yasg',
```
nemesisdesign, line 57: great :+1:, please add a mention in the README about how to see the API docs (add a screenshot please, it will help ... and in the future we'll be adding a lot more screenshots to the READMEs of each module).

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,80 @@
...
+            if org:
+                organization_filter = {self.organization_field + '__name': org}
+                queryset = queryset.filter(**organization_filter)
+        except Exception:
+            # django.core.exceptions.ValidationError when uuid is not valid
+            queryset = []
+        return queryset
+
+
+class BuildAPIListView(OrgAPIMixin, generics.ListCreateAPIView):
```
nemesisdesign, line 35: I'd remove `API` from the name of the views since we're already in `api.py` and it's implicit.

openwisp_firmware_upgrader/api/tests/base/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+from swapper import load_model
+
+from openwisp_users.tests.utils import TestMultitenantAdminMixin
+
+BatchUpgradeOperation = load_model('firmware_upgrader', 'BatchUpgradeOperation')
+Build = load_model('firmware_upgrader', 'Build')
+Category = load_model('firmware_upgrader', 'Category')
+DeviceFirmware = load_model('firmware_upgrader', 'DeviceFirmware')
+FirmwareImage = load_model('firmware_upgrader', 'FirmwareImage')
+UpgradeOperation = load_model('firmware_upgrader', 'UpgradeOperation')
```
nemesisdesign, line 14: do not load the concrete models here, this may cause issues because this file will be imported also by extension apps (eg: the sample app)

openwisp_firmware_upgrader/api/tests/base/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+user_model = get_user_model()
+
+
+class TestAPIUpgraderMixin(TestMultitenantAdminMixin, TestUpgraderMixin):
+    device_firmware_model = DeviceFirmware
+    upgrade_operation_model = UpgradeOperation
+    batch_upgrade_operation_model = BatchUpgradeOperation
+    firmware_image_model = FirmwareImage
+    build_model = Build
+    category_model = Category
```
nemesisdesign, line 26: if it's a mixin, it can live also without the definitions of the models here, that definition has to be put in the test class which inherits the mixin

openwisp_firmware_upgrader/api/tests/base/__init__.py
PabloCastellano: I've moved them back to test_views.py, let me know if this is the correct place

openwisp_firmware_upgrader/api/urls.py
```diff
@@ -0,0 +1,63 @@
...
+            schema_view.with_ui('swagger', cache_timeout=0),
+            name='schema-swagger-ui',
+        ),
+        url(
+            r'^redoc/$',
+            schema_view.with_ui('redoc', cache_timeout=0),
+            name='schema-redoc',
+        ),
+        # API endpoints
+        path('build/', views.build_list, name='api-build-list'),
```
nemesisdesign, line 45: let's switch to underscores in view names

openwisp_firmware_upgrader/api/urls.py
```diff
@@ -0,0 +1,63 @@
...
+        path('build/<pk>/', views.build_detail, name='api-build-detail'),
+        path('firmware/', views.firmware_image_list, name='api-firmware-list',),
```
nemesisdesign, line 47: can we have something like:

```
build/<pk>/firmware-image/
```

Which lists the firmware images of the build?
Listing all the images of all the builds will not be very useful.

openwisp_firmware_upgrader/api/urls.py
```diff
@@ -0,0 +1,63 @@
...
+        path(
+            'firmware/<pk>/', views.firmware_image_detail, name='api-firmware-detail',
```
nemesisdesign, line 49: if we can be consistent with the previous requested change, hence having the following, it would be great:

```
build/<pk>/firmware-image/<pk>/
```

openwisp_firmware_upgrader/api/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+class CategorySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Category
+        fields = '__all__'
+
+
+class BatchUpgradeOperationSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = BatchUpgradeOperation
+        fields = '__all__'
```
nemesisdesign, line 25: we need a nested serializer (readonly) in which we list the related operations (as in the admin), this is needed to monitor the progress of a mass upgrade

PabloCastellano: Cool. Should I add a nested serializer for `Build` or do we want to have also fields like `completed`, `success_rate`, ...?

nemesisdesign: Yes I think we should add a nested serializer and I also think it would be good to have `sucess_rate` and similar info in this view, it should be possible to accomplish it by adding specific methods to the serializer class.

openwisp_firmware_upgrader/api/urls.py
PabloCastellano: I also thought about it.

Do you think we should we keep `/firmware-image/` then or just your `build/<pk>/firmware-image/`?

nemesisdesign: a readonly endpoint with some filters may be useful

.travis.yml
```diff
@@ -33,11 +33,13 @@ before_install:
...
 
 install:
-  - pip install -e .
+  - pip install -e .[rest]
   # temporary: remove when openwisp-controller 0.7.0 is released
   - pip install -U https://github.com/openwisp/openwisp-controller/tarball/master
   # temporary: remove when openwisp-utils 0.5.0 is released
   - pip install -U https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils[qa]
+  # temporary: remove when openwisp-users with API support is released
+  - pip install -U https://github.com/PabloCastellano/openwisp-users/tarball/issue/104#egg=openwisp_users[rest]
```
nemesisdesign, line 42: you can now use the master branch

openwisp_firmware_upgrader/tests/test_api/test_views.py
```diff
@@ -0,0 +1,534 @@
...
+    firmware_image_model = FirmwareImage
+    upgrade_operation_model = UpgradeOperation
+
+    def _serialize_build(self, build):
+        serializer = BuildSerializer()
+        return dict(serializer.to_representation(build))
+
+    def test_build_unauthorized(self):
+        build = self._create_build()
+        client = Client()
```
nemesisdesign, line 34: these tests should go in `tests/tst_api/base/test_views.py` and be imported here as well as in the sample_app tests.
Applies for the other test classes as well.

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+    queryset = Build.objects.all()
+    serializer_class = BuildSerializer
+    organization_field = 'category__organization'
+
+
+class BuildDetailView(ProtectedAPIMixin, generics.RetrieveUpdateDestroyAPIView):
+    queryset = Build.objects.all()
+    serializer_class = BuildSerializer
+    lookup_fields = ['pk']
+    permission_classes = [IsAuthenticated]
```
nemesisdesign, line 53: what happens if somebody tries to access a build which is related to an organization the user does not have access to?

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+    queryset = Category.objects.all()
+    serializer_class = CategorySerializer
+    organization_field = 'organization'
+    permission_classes = [IsAuthenticated]
+
+
+class CategoryDetailView(ProtectedAPIMixin, generics.RetrieveUpdateDestroyAPIView):
+    queryset = Category.objects.all()
+    serializer_class = CategorySerializer
+    lookup_fields = ['pk']
```
nemesisdesign, line 66: same question regarding the organization here

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+):
+    queryset = BatchUpgradeOperation.objects.all()
+    serializer_class = BatchUpgradeOperationSerializer
+    organization_field = 'build__category__organization'
+
+
+class BatchUpgradeOperationDetailView(ProtectedAPIMixin, generics.RetrieveAPIView):
+    queryset = BatchUpgradeOperation.objects.all()
+    serializer_class = BatchUpgradeOperationSerializer
+    lookup_fields = ['pk']
```
nemesisdesign, line 80: same question regarding the organization here

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+        self.queryset = FirmwareImage.objects.filter(build=build_pk)
+        queryset = super().get_queryset()
+        return queryset
+
+    def create(self, request, *args, **kwargs):
+        request.data['build'] = kwargs['pk']
+        return super().create(request, *args, **kwargs)
+
+
+class FirmwareImageDetailView(ProtectedAPIMixin, generics.RetrieveUpdateDestroyAPIView):
```
nemesisdesign, line 98: same question regarding the organization here

tests/openwisp2/firmware/fake-img.bin
```diff
@@ -0,0 +1 @@
+dummy
```
nemesisdesign, line 1: haha, can you believe the cause of the test failures is this line?? :joy: :joy:

PabloCastellano: OMG :man_facepalming: this is somewhat required otherwise uploading an empty firmware image will fail

nemesisdesign: why will it fail? These tests were passing fine before.

PabloCastellano: because you cannot upload an empty file

nemesisdesign: @PabloCastellano I'd create a new file and leave that one alone.

nemesisdesign: And I'd add a comment in the new file explaining it's needed to test the API and it cannot be empty.

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+)
+
+BatchUpgradeOperation = load_model('firmware_upgrader', 'BatchUpgradeOperation')
+Build = load_model('firmware_upgrader', 'Build')
+Category = load_model('firmware_upgrader', 'Category')
+FirmwareImage = load_model('firmware_upgrader', 'FirmwareImage')
+
+
+class ProtectedAPIMixin(object):
+    authentication_classes = [BearerAuthentication]
```
nemesisdesign, line 21: can you add `SessionAuthentication`? It will make it easier for us to test the API via the browsable API interface.

openwisp_firmware_upgrader/api/serializers.py
nemesisdesign: hey, I see that we have a nested serializer for the build object, that's good, but we also need a nested serializer which lists all the upgrade operations, is this clear?

PabloCastellano: I see, it's clearer now :+1:

openwisp_firmware_upgrader/api/views.py
PabloCastellano: Well spotted. I've fixed it and now returns 404 if the user does not belong to the organization

openwisp_firmware_upgrader/api/views.py
PabloCastellano: Fixed too, and added test cases for Category and the rest of endpoints

openwisp_firmware_upgrader/api/views.py
```diff
@@ -0,0 +1,117 @@
...
+    permission_classes = [IsAuthenticated]
```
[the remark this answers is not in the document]
PabloCastellano: Right, fixed

PabloCastellano:
Addressed all requested changes :tada:

PabloCastellano:
@nemesisdesign I have tried another approach: using [ViewSets](https://www.django-rest-framework.org/api-guide/viewsets/). It reduces the amount of code in views.py and all tests keep passing. The only difference is that it expects url names with dashes instead of underscores and changing it would probably be against the initial reduction of amount of code. WDYT? Do we prefer keeping underscores over dashes?

PabloCastellano:
@nemesisdesign ScopedRateThrottling looks like the way to go. We can change it in openwisp-users for sure. Do we want to rate limit views in openwisp-firmware-upgrader too?

PR #51, opened 2020-05-09
[api] Make firmware image not editable #47

### LINKED ISSUE #47: [api] Make firmware image not editable
@PabloCastellano see this: https://github.com/openwisp/openwisp-firmware-upgrader/issues/44

I think we can make the firmware images not editable.
Users will have to delete and recreate it.

This makes several things easier and also makes sense since we want to avoid the situation in which all devices are flashed with an image, then the image file is changed with a different one but the system still believes the devices are using that image (they're not, they'll be using a different one).
So forcing the users to delete the image if they want to change it, also makes the system aware of this.

I think we can do this in the API (we'll do it also in the admin in #44).

This means you can remove this:

https://github.com/openwisp/openwisp-firmware-upgrader/blob/f024859288397f763debd4c9806ca8f82ab7ba23/openwisp_firmware_upgrader/tests/base/test_api.py#L574-L590

And its related functionality.

PabloCastellano:
Closes #47

### DIFF
--- a/openwisp_firmware_upgrader/api/urls.py
+++ b/openwisp_firmware_upgrader/api/urls.py
@@ -1,5 +1,5 @@
 from django.conf.urls import url
-from django.urls import path
+from django.urls import include, path
 from drf_yasg import openapi
 from drf_yasg.views import get_schema_view
 from openwisp_firmware_upgrader import settings as app_settings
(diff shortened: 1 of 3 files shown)


openwisp_firmware_upgrader/api/urls.py
```diff
@@ -43,37 +43,44 @@
...
-        path(
-            'firmware/batch-upgrade-operation/',
-            views.batch_upgrade_operation_list,
-            name='api_batchupgradeoperation_list',
-        ),
-        path(
-            'firmware/batch-upgrade-operation/<pk>/',
-            views.batch_upgrade_operation_detail,
-            name='api_batchupgradeoperation_detail',
+            'firmware/',
```
nemesisdesign, line 47: this would be one of those changes you can put in its own commit and can stay like that

openwisp_firmware_upgrader/api/views.py
```diff
@@ -119,7 +119,7 @@ def create(self, request, *args, **kwargs):
         return super().create(request, *args, **kwargs)
 
 
-class FirmwareImageDetailView(OrgAPIMixin, generics.RetrieveUpdateDestroyAPIView):
+class FirmwareImageDetailView(OrgAPIMixin, generics.RetrieveDestroyAPIView):
```
nemesisdesign, line 122: looks good

PR #53, opened 2020-05-09
[tests] Assert number of queries in API calls #50

### LINKED ISSUE #50: [api] Keep the number of queries low
The number of queries can skyrocket if left unchecked.

We should add assertions in the tests to check this.

We'll reuse the same technique across all the modules where we introduce an API.

https://docs.djangoproject.com/en/dev/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries

PabloCastellano:
Closes #50


PabloCastellano:
@nemesisdesign The number of queries differ when `SAMPLE_APP=1` is specified. How do we deal with it?

PabloCastellano:
This is weird. This gist contains the sql queries made. I'm investigating but have no clue so far

https://gist.github.com/PabloCastellano/95ec12f684343f8093734a6de1cb3e2f

PabloCastellano:
Ok I found it!

Check out this code from DRF:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/fields.py#L576-L602

Line 589 is where the each validator is run.

It turns out that the Category and FirmwareImage models defined with `SAMPLE_APP=1` don't have the Meta.unique_together attribute

tests/openwisp2/sample_firmware_upgrader/migrations/0004_models_unique_together.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='build', unique_together={('category', 'version')},
+        ),
+        migrations.AlterUniqueTogether(
+            name='category', unique_together={('name', 'organization')},
+        ),
+        migrations.AlterUniqueTogether(
+            name='firmwareimage', unique_together={('build', 'type')},
```
nemesisdesign, line 21: [given enough eyeballs, all bugs are shallow](https://en.wiktionary.org/wiki/given_enough_eyeballs,_all_bugs_are_shallow) :+1:

PR #54, opened 2020-05-09
[api] Move swagger to openwisp-utils #45

### LINKED ISSUE #45: [api] Move swagger to openwisp-utils
Move swagger to openwsp-utils as indicated here: https://github.com/openwisp/openwisp-utils/issues/102
And then update the code of this module to use that one instead.

PabloCastellano:
Fixes #45

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,9 +38,6 @@ setup(
         'django-private-storage~=2.2',
         'swapper~=1.1',
     ],
-    extras_require={
-        'rest': ['djangorestframework>=3.11,<3.12', 'drf-yasg>=1.17,<1.18']
-    },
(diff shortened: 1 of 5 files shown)


PabloCastellano:
I don't see why coverage has decreased

setup.py
```diff
@@ -38,9 +38,6 @@
         'django-private-storage~=2.2',
         'swapper~=1.1',
     ],
-    extras_require={
-        'rest': ['djangorestframework>=3.11,<3.12', 'drf-yasg>=1.17,<1.18']
```
nemesisdesign, line 42: I think here we need to put `openwisp-utils[rest]`

PR #57, opened 2020-05-13
 [api] User spanning multiple orgs crashes the API #55

### LINKED ISSUE #55: [api] User spanning multiple orgs crashes the API
Here's the issue:

https://github.com/openwisp/openwisp-firmware-upgrader/blob/b1238a83ea1a40a60261b9a57a0e1a23debda6e7/openwisp_firmware_upgrader/api/views.py#L34-L37

Instead of expecting one org, we should expect multiple ones, the query should be something like `organization__in=[org1_uuid, org2_uuid]`

**PS**: we need to write some tests using an user which is not superuser, it seems this use case is broken.

PabloCastellano:
Fixes #55

### DIFF
--- a/openwisp_firmware_upgrader/api/views.py
+++ b/openwisp_firmware_upgrader/api/views.py
@@ -31,10 +31,10 @@ class ProtectedAPIMixin(object):
 
 class OrgAPIMixin(ProtectedAPIMixin):
     def get_queryset(self):
-        queryset = self.queryset.all().order_by('-pk')
+        queryset = self.queryset.all()
         if not self.request.user.is_superuser:
             user_orgs = self.request.user.openwisp_users_organization.all()
-            filter_key = self.organization_field + "__in"
+            filter_key = f'{self.organization_field}__in'
             organization_filter = {filter_key: user_orgs}
             queryset = queryset.filter(**organization_filter)
         org = self.request.query_params.get('org', None)
--- a/openwisp_firmware_upgrader/api/views.py
+++ b/openwisp_firmware_upgrader/api/views.py
@@ -33,8 +33,9 @@ class OrgAPIMixin(ProtectedAPIMixin):
     def get_queryset(self):
         queryset = self.queryset.all().order_by('-pk')
         if not self.request.user.is_superuser:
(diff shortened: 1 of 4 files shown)


openwisp_firmware_upgrader/api/views.py
```diff
@@ -33,8 +33,9 @@ class OrgAPIMixin(ProtectedAPIMixin):
     def get_queryset(self):
         queryset = self.queryset.all().order_by('-pk')
         if not self.request.user.is_superuser:
-            user_org = self.request.user.openwisp_users_organization.get()
-            organization_filter = {self.organization_field: user_org}
+            user_orgs = self.request.user.openwisp_users_organization.all()
+            filter_key = self.organization_field + "__in"
```
nemesisdesign, line 37: I'd change to:

```python
filter_key = f'{self.organization_field}__in'
```

(for consistency and is more pythonic).

openwisp_firmware_upgrader/api/views.py
```diff
@@ -33,8 +33,9 @@ class OrgAPIMixin(ProtectedAPIMixin):
     def get_queryset(self):
         queryset = self.queryset.all().order_by('-pk')
         if not self.request.user.is_superuser:
-            user_org = self.request.user.openwisp_users_organization.get()
-            organization_filter = {self.organization_field: user_org}
+            user_orgs = self.request.user.openwisp_users_organization.all()
```
nemesisdesign, line 36: I took a note to remind us to cache this: https://github.com/openwisp/openwisp-firmware-upgrader/issues/59

openwisp_firmware_upgrader/api/views.py
```diff
@@ -33,8 +33,9 @@ class OrgAPIMixin(ProtectedAPIMixin):
     def get_queryset(self):
         queryset = self.queryset.all().order_by('-pk')
```
nemesisdesign, line 34: that `.order_by('-pk')` does not make a lot of sense (because we use UUIDs which are not sequential so the order would be random). Moreover, we also specified the default ordering recently for the endpoints. I think it can be removed.

PR #58, opened 2020-05-15
[api] Implement filters

### LINKED ISSUE #49: [api] Filters
It should be able to replace the org filter with https://www.django-rest-framework.org/api-guide/filtering/#djangofilterbackend which should result cleaner and also will help us to implement the same feature faster when we'll add the API in the other modules.

If adding filters with that method is simple as it seems we could add a couple more:

- filter builds by category
- filter batch upgrade operations by status

PabloCastellano:
Implement filters for Build, FirmwareImage and BatchUpgradeOpeation

Requires django-filters

* BatchUpgradeOperations can be filtered by status and build
* Builds can be filtered by category
* FirmwareImages can be filtered by type

Closes #49


PabloCastellano:
According to django-filters documentation, I had to add `django_filters` to `INSTALLED_APPS`, however I found that this works fine without adding it, so I haven't done it

PR #60, opened 2020-05-18
Swappable improvements

nemesisdesign:
I've been wondering: do we really have to keep separate abstract admin and test classes?

We had to do that when we were not using swappable models, because otherwise concrete models would be loaded in custom apps, which would trigger exceptions by django, since those concrete models were not part of `INSTALLED_APPS`.

But now that we use swappable models, the concrete models are never imported directly, but always loaded through `swappable.load_model`, which means the work we're doing to create abstract admin and abstract test classes is redundant.

I think this PR simplifies things for maintainers and it would be beneficial to adopt the same method in all the other modules.

Let me know what you think @atb00ker.

PS: regarding the commit `[swappable] Added shortcut functions`, I thought it was a good idea and realized only later that in many modules we have different django apps (see openwisp-controller, openwisp-monitoring), so I wouldn't enforce doing that but leave it to the preference of the contributor.

PPS: in order to be sure the tests for the sample app are actually doing anything useful, try editing DetailsModel as follows:

```python
class DetailsModel(models.Model):
    details = models.CharField(max_length=64, blank=True, null=True)

    def clean(self):
        raise ValueError('see how sample tests fail?')

    class Meta:
        abstract = True
```

Then run:

```
./runtests.py --keepdb --parallel
SAMPLE_APP=1 ./runtests.py --keepdb --parallel
```

You should see most of the tests of the second command fail, while the tests in the first command will all pass.


tests/openwisp2/sample_firmware_upgrader/tests.py
```diff
@@ -1,108 +1,55 @@
...
-    upgrade_operation_model = UpgradeOperation
-    batch_upgrade_operation_model = BatchUpgradeOperation
-    firmware_image_model = FirmwareImage
-    build_model = Build
-    category_model = Category
+class TestTasks(BaseTestTasks):
+    pass
+
+
+# this is necessary to avoid excuting the base test suites
```
PabloCastellano, line 49: This part looks very hacky. What if the base classes don't just inherit from TestCase?

E.g.:

```diff
# openwisp_firmware_upgrader/tests/test_models.py
-class TestModels(TestUpgraderMixin, TestCase):
+class TestModels(TestUpgraderMixin):
    app_label = 'openwisp_firmware_upgrader'
   ...
```

and then

```diff
# tests/openwisp2/sample_firmware_upgrader/tests.py
-class TestModels(BaseTestModels):
+class TestModels(BaseTestModels, TestCase):
    app_label = 'openwisp2.sample_firmware_upgrader'
```

tests/openwisp2/sample_firmware_upgrader/admin.py
```diff
@@ -1,26 +1,5 @@
-from django.contrib import admin
-from openwisp_firmware_upgrader.base.admin import (
-    AbstractBatchUpgradeOperationAdmin,
-    AbstractBuildAdmin,
-    AbstractCategoryAdmin,
+from openwisp_firmware_upgrader.admin import (  # noqa
```
atb00ker, line 1: I think we should mention in README to unregister and register a class again in case you want to make changes to an admin class, right?

tests/openwisp2/sample_firmware_upgrader/tests.py
nemesisdesign: I agree is hacky.

The solution proposed would result in tests not being executed when runnning: `./runtests.py`
Only `SAMPLE_APP=1 ./runtests.py` would run actual tests.

The alternative would be:

```diff
# openwisp_firmware_upgrader/tests/test_models.py
-class TestModels(TestUpgraderMixin, TestCase):
+class BaseTestModels(TestUpgraderMixin):
    app_label = 'openwisp_firmware_upgrader'
   ...


+class TestModels(BaseTestModels):
+    pass
```

Which is somewhat less hacky but still weird, so I thought that if we have to have weird code (since there's no way around it), having it in the sample app is better since we deal with it a lot less (we work far more often on the base tests than the sample app tests) the weird lines are concentrated in one place, this is inline with the goal of making it simpler for maintainers to work on it I think, what do you think about this?

tests/openwisp2/sample_firmware_upgrader/admin.py
nemesisdesign: If they want to make big changes, yes, it would be better for them to unregister, extend the class and register it again, for small changes they can resort to monkey patching.

nemesisdesign: I am thinking about further improving this, will push shortly.

nemesisdesign: For a moment I thought that we could check if the main app is in `INSTALLED_APPS` and if not, we can avoid registering the admin classes.

But with the information I have now, the amount of people needing this is not high enough to justify the effort.

I think unregistering and registering the admin classes is good enough.

nemesisdesign:
@atb00ker @nepython @pandafy please see my latest commits in this PR and replicate these practices in your swappable implementation

tests/openwisp2/sample_firmware_upgrader/admin.py
atb00ker: One possible easy solution can be: 
```
example_model = swapper.load_model("example")
if not admin.is_registered(example_model):
    admin.register(example_adminclass, example_model)
```

tests/openwisp2/sample_firmware_upgrader/tests.py
```diff
@@ -1,108 +1,55 @@
-import os
-from unittest import skipUnless
-
-from django.test import TestCase, TransactionTestCase
 from django.urls import reverse
```
atb00ker, line 1: In the openwisp-ipam, I tried:
```
from openwisp_ipam.tests.test_admin import TestAdmin
from openwisp_ipam.tests.test_api import TestApi

# Monkey Patching to change the name of the app_label
# used tests. Change this value to your <app_label>.
TestAdmin.app_label = 'sample_ipam'
```

and it worked, isn't that simpler or Is there a reason you've avoided that here?

README.rst
```diff
@@ -312,40 +312,72 @@ this concept, read the following section.
...
-The ``tests/openwisp2/sample-firmware-upgrader`` may serve as an example for
-extending *openwisp-firmware-upgrader* in your own application.
+One of the core values of the OpenWISP project is `Software Reusability <http://openwisp.io/docs/general/values.html#software-reusability-means-long-term-sustainability>`_,
+for this reason *openwisp-firmware-upgrader* provides a set of base classes
+which can be imported, extended and reused to create derivative apps.
 
-*openwisp-firmware-upgrader* provides a set of models and admin classes which can
-be imported, extended and reused by third party apps.
+In order to implement your custom version of *openwisp-firmware-upgrader*,
+you need to perform the steps described in the rest of this section.
```
atb00ker, line 320: `in the rest of this section` -> `in this section`

README.rst
```diff
@@ -312,40 +312,72 @@ this concept, read the following section.
...
 
-To extend *openwisp-firmware-upgrader*, **you MUST NOT** add it to ``settings.INSTALLED_APPS``,
-but you must create your own app (which goes into ``settings.INSTALLED_APPS``), import the
-base classes from openwisp-firmware-upgrader and add your customizations.
-
-In order to help django find the static files and templates of *openwisp-firmware-upgrader*,
-you need to perform the steps described below.
+When in doubt, the code in the `test project <https://github.com/openwisp/openwisp-firmware-upgrader/tree/master/tests/openwisp2/>`_
```
atb00ker, line 322: Also, what do you think about using relative path for these links? (`tests/openwisp2` in this case) 
So that these links work in fork and other branches as well, absolute path will work only in master branch!

README.rst
```diff
@@ -385,51 +417,209 @@ Add ``openwisp_utils.loaders.DependencyLoader`` to ``TEMPLATES`` in your ``setti
...
+Create and apply database migrations::
+
+    ./manage.py makemigrations
+    ./manage.py migrate
+
+For more information, refer to the
+`"Migrations" section in the django documentation <https://docs.djangoproject.com/en/dev/topics/migrations/>`_.
+
+10. Create the admin
 ~~~~~~~~~~~~~~~~~~~
```
atb00ker, line 473: Additional `~` required here for making it equal length with `10. Create the admin` :smile:

README.rst
```diff
@@ -385,51 +417,209 @@ Add ``openwisp_utils.loaders.DependencyLoader`` to ``TEMPLATES`` in your ``setti
...
+    from openwisp_firmware_upgrader.admin import (  # noqa
+        BatchUpgradeOperationAdmin,
+        BuildAdmin,
+        CategoryAdmin,
+    )
+
+    BuildAdmin.list_display.insert(1, 'my_custom_field')
+    BuildAdmin.ordering = ['-my_custom_field']
+
+2. Inheriting admin classes
```
atb00ker, line 500: While, doing this section in openwisp-ipam, I thought about focusing only on one model instead of showing all of them for brevity.

Added the line `@admin.register(IpAddress)`; What do you think about that?

Final result:
```
If you need to introduce significant changes and/or you don't want to resort to
monkey patching, you can proceed as follows, here we make changes in the IpAddressAdmin:

.. code-block:: python
    from django.contrib import admin
    from openwisp_ipam.admin import IpAddressAdmin as BaseIpAddressAdmin
    from swapper import load_model

    IpAddress = load_model('openwisp_ipam', 'IpAddress')

    admin.site.unregister(IpAddress)

    @admin.register(IpAddress)
    class IpAddressAdmin(BaseIpAddressAdmin):
        # add your changes here
```

README.rst
```diff
@@ -385,51 +417,209 @@ Add ``openwisp_utils.loaders.DependencyLoader`` to ``TEMPLATES`` in your ``setti
...
+
+13. Import the automated tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When developing a custom application based on this module, it's a good
-idea to import and run the `base tests <https://github.com/openwisp/openwisp-firmware-upgrader/tree/master/openwisp_firmware_upgrader/tests/base>`_
-too, so that you can be sure the changes you're introducing are not breaking
-some of the existing features of openwisp-firmware-upgrader.
+idea to import and run the base tests too, so that you can be sure the changes
+you're introducing are not breaking some of the existing features of openwisp-firmware-upgrader.
```
atb00ker, line 559: Just to point out a small detail, everywhere else, you've used `*openwisp-firmware-upgrader*`, here it's only `openwisp-firmware-upgrader`. :smile:

tests/openwisp2/sample_firmware_upgrader/tests.py
atb00ker: Okay, adding tests would require monkey patching -- dirty solution, I am assuming that's why It's not used. :smile:

tests/openwisp2/sample_firmware_upgrader/admin.py
nemesisdesign: this where, in the main admin.py file?

README.rst
nemesisdesign: Would relative links from both type of URLs?

- https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/README.rst
- https://github.com/openwisp/openwisp-firmware-upgrader/

I think it would either work in one but not in the other.

tests/openwisp2/sample_firmware_upgrader/tests.py
nemesisdesign: I did not think about that. It It came natural to do it in the other way. I think both solutions are a bit hacky, it's just a matter of choice. Which one gives you less pain when you look at it? HEHE

README.rst
nemesisdesign: Have you ever been a coach for events like Djangogirls?
I've been a few times.
People read the docs and while they do they copy the examples verbatim and paste them into their editors.
We do it as well right? I think we should do a bit of effort to make it easier for them.
We can sacrifice brevity for the sake of completeness.
For you it's obvious that the same pattern can be applied to other models, but to someone which knows nothing about django and about the module is approaching, may be totally obscure.

atb00ker: Okay, in thatcase, the only change is adding information about registration `@admin.register(IpAddress)`

tests/openwisp2/sample_firmware_upgrader/admin.py
atb00ker: I tested it in openwisp_ipam, it wouldn't work! :-(

tests/openwisp2/sample_firmware_upgrader/tests.py
atb00ker: Fair enough! :smile:
I've also implement this one because it'll be easier for people not so well versed with python. :smile:

README.rst
atb00ker: @nemesisdesign sorry, I didn't get that! :smile:

README.rst
```diff
@@ -385,51 +417,209 @@ Add ``openwisp_utils.loaders.DependencyLoader`` to ``TEMPLATES`` in your ``setti
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Please refer to the `urls.py <https://github.com/openwisp/openwisp-firmware-upgrader/tree/master/tests/openwisp2/urls.py>`_
+file in the test project.
+
+For more information about URL configuration in django, please refer to the
+`"URL dispatcher" section in the django documentation <https://docs.djangoproject.com/en/dev/topics/http/urls/>`_.
+
+12. Create celery.py
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
atb00ker, line 546: Just noticed this: 
```
12. Create celery.py
~~~~~~~~~~~~~~~~
```
`~~~~~~~~~~~~~~~~~` length changed! :smile:

nemesisdesign: Addressed, thanks

nemesisdesign:
I think is good enough, I'm merging. For subsequent improvements we'll do on master. Keep discussing doubts here if needed.

README.rst
nemesisdesign: The README can be viewed on two different URLs (the main repo URL and /blob/master/README.rst), which are on different levels, so using a relative path would work in one but not in the other.

tests/openwisp2/sample_firmware_upgrader/tests.py
PabloCastellano: Fair enough

PR #64, opened 2020-05-21
[helpers] Use new User.is_member() in FirmwareImageDownloadView

PabloCastellano:
Related to https://github.com/openwisp/openwisp-users/issues/107


openwisp_firmware_upgrader/private_storage/views.py
```diff
@@ -13,9 +13,10 @@ def can_access_file(self, private_file):
         user = private_file.request.user
         if not user.is_authenticated:
             return False
+        if user.is_superuser:
+            return True
         org = self.object.build.category.organization
-        user_organizations = user.openwisp_users_organization.all()
-        return user.is_superuser or (user.is_staff and org in user_organizations)
+        return user.is_staff and user.is_member(org)
```
nemesisdesign, line 19: I think we should use `is_manager()`, because `is_member()` is meant for end-users, eg: public wifi users or in any case, users which just consume some kind of service but do not manage the firmware images, so they should not be able to download them.

Does this work the same in the API endpoint to download firmware images?

I was thinking, can you call `private_storage.FirmwareImageDownloadView` from `api.views.FirmwareImageDownloadView`?

```python
private_storage.FirmwareImageDownloadView.as_view(request, path)
```

This would allow us to keep the behaviour in the API and the admin in sync.

If you think you can do it, could you please open a new issue for this and assign it to yourself?

PabloCastellano: I've made the changes but the test is failing now because `AttributeError: This FileResponse instance has no `content` attribute. Use `streaming_content` instead.`

Also the number of queries has increased from 2 to 5.

Let me know what you think

openwisp_firmware_upgrader/tests/test_api.py
```diff
@@ -29,6 +29,7 @@ class TestAPIUpgraderMixin(TestMultitenantAdminMixin, TestUpgraderMixin):
     def setUp(self):
         self.org = self._get_org()
         self.operator = self._create_operator(organizations=[self.org])
+        self.operator.organizations_dict  # force caching
```
PabloCastellano, line 32: What do you think about this hack? I did it because using organizations_dict in OrgAPIMixin, makes an extra query than previously. I haven't come down to the reason.

Do you know if there was some cache involved?

```diff
-            user_orgs = self.request.user.openwisp_users_organization.all()
+            user_orgs = self.request.user.organizations_dict.keys()
```

PR #72, opened 2020-05-29
Minor improvements in admin and docker-compose

PabloCastellano:
* [admin] Don't show the cancel button when there are no operations
* [docker] Add docker-compose for Redis

### DIFF
--- a/openwisp_firmware_upgrader/templates/admin/upgrade_selected_confirmation.html
+++ b/openwisp_firmware_upgrader/templates/admin/upgrade_selected_confirmation.html
@@ -52,7 +52,11 @@
   <input type="submit" name="upgrade_related" value="{% trans "Upgrade only related devices" %}">
   {% endif %}
   {% endif %}
+  {% if related_count or firmwareless_count %}
   <a href="#" class="button cancel-link">{% trans "Cancel the upgrade" %}</a>
+  {% else %}
+  <a href="#" class="button cancel-link">{% trans "Go back" %}</a>
+  {% endif %}
   </div>
   </form>
 {% endblock %}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,8 @@
+version: "3"
+
+services:
+  redis:
+    image: redis:alpine
+    ports:
+      - "6379:6379"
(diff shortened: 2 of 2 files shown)


openwisp_firmware_upgrader/admin.py
```diff
@@ -102,9 +102,9 @@ def upgrade_selected(self, request, queryset):
             {
                 'title': title,
                 'related_device_fw': related_device_fw,
-                'related_count': len(related_device_fw),
+                'related_count': related_device_fw.count(),
                 'firmwareless_devices': firmwareless_devices,
-                'firmwareless_count': len(firmwareless_devices),
+                'firmwareless_count': firmwareless_devices.count(),
```
nemesisdesign, line 107: both querysets would be evaluated anyway during template rendering to build the list of devices, hence the full query will be done anyway.
So with this change, we're adding 2 more queries.
Or am I missing something?

nemesisdesign: As anticipated, I added an assertion for the number of queries in the tests to test whether my assumption was correct.

PabloCastellano: Well spotted, I've removed that part

openwisp_firmware_upgrader/templates/admin/upgrade_selected_confirmation.html
```diff
@@ -52,7 +52,9 @@ <h2>{% trans "Devices which the system has never upgraded yet" %} ({{ firmwarele
   <input type="submit" name="upgrade_related" value="{% trans "Upgrade only related devices" %}">
   {% endif %}
   {% endif %}
+  {% if related_count or firmwareless_count %}
```
nemesisdesign, line 55: So if there's no devices, how do the users go back? Maybe we could still provide the button but change its anchor text to show something like:

```
No devices to upgrade

<link>Go back</link>
```

PabloCastellano: In my case I just clicked on the back button of my browser but I like the idea of having an extra button saying "Go back"

PR #80, opened 2020-06-03
[admin] Show in progress upgrade operation in Device Firmware tab #65

### LINKED ISSUE #65: [admin] Show in progress upgrade operation in Device Firmware tab
Not super obvious, but it's possible to launch single upgrade operations for each device from the device page, using the Device Firmware tab (selecting a firmware image and saving).

First of all, we should order the images so the latest images come first.
We should also filter the image to show only the images for the type of hardware being used in that device (if the information is available).

However if the single upgrade opration is launched this way, there's no way to track its progress.

A simple solution is to look if an UpgradeOperation related to that device is in progress and if yes, show its details there.

PabloCastellano:
* [ ] Fix tests (not sure why they pass in my machine and fail in Travis)

Closes #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ python:
   - "3.7"
 
 env:
-  - DJANGO="django>=2.2,<2.3"
   - DJANGO="django>=3.0,<3.1"
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,6 @@ before_install:
 install:
   # temporary: remove when openwisp-controller 0.7.0 is released
   - pip install -U https://github.com/openwisp/openwisp-controller/tarball/master
-  # temporary: remove when openwisp-utils 0.5.0 is released
-  - pip install -U https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils[qa,rest]
   # temporary: remove when openwisp-users with API support is released
   - pip install -U https://github.com/openwisp/openwisp-users/tarball/master#egg=openwisp_users[rest]
   - pip install -e .
--- a/openwisp_firmware_upgrader/admin.py
+++ b/openwisp_firmware_upgrader/admin.py
@@ -1,12 +1,12 @@
 import logging
+from datetime import timedelta
 
-from dateutil.relativedelta import relativedelta
 from django import forms
 from django.contrib import admin, messages
 from django.contrib.admin.helpers import ACTION_CHECKBOX_NAME
 from django.shortcuts import redirect
 from django.template.response import TemplateResponse
-from django.urls import reverse
+from django.urls import resolve, reverse
 from django.utils.safestring import mark_safe
 from django.utils.timezone import localtime
 from django.utils.translation import ugettext_lazy as _
@@ -16,12 +16,14 @@ from openwisp_controller.config.admin import DeviceAdmin
 from openwisp_users.multitenancy import MultitenantAdminMixin
 from openwisp_utils.admin import ReadOnlyAdmin, TimeReadonlyAdminMixin
 
-# from .hardware import FIRMWARE_IMAGE_MAP
+from .hardware import REVERSE_FIRMWARE_IMAGE_MAP
 from .swapper import load_model
 
 logger = logging.getLogger(__name__)
 BatchUpgradeOperation = load_model('BatchUpgradeOperation')
 UpgradeOperation = load_model('UpgradeOperation')
+DeviceFirmware = load_model('DeviceFirmware')
+FirmwareImage = load_model('FirmwareImage')
 
 
 class BaseAdmin(MultitenantAdminMixin, TimeReadonlyAdminMixin, admin.ModelAdmin):
@@ -42,7 +44,7 @@ class CategoryAdmin(BaseVersionAdmin):
 
 
 class FirmwareImageInline(TimeReadonlyAdminMixin, admin.StackedInline):
-    model = load_model('FirmwareImage')
+    model = FirmwareImage
     extra = 0
 
     def has_change_permission(self, request, obj=None):
@@ -139,7 +141,7 @@ class UpgradeOperationForm(forms.ModelForm):
 
 
 class UpgradeOperationInline(admin.StackedInline):
-    model = load_model('UpgradeOperation')
+    model = UpgradeOperation
     form = UpgradeOperationForm
     readonly_fields = UpgradeOperationForm.Meta.fields
     extra = 0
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -40,6 +40,8 @@ install:
   - pip install -U https://github.com/openwisp/openwisp-utils/tarball/master#egg=openwisp_utils[qa,rest]
   # temporary: remove when openwisp-users with API support is released
   - pip install -U https://github.com/openwisp/openwisp-users/tarball/master#egg=openwisp_users[rest]
+  # temporary:
+  - pip install celery==4.4.2
```
nemesisdesign, line 44: shouldn't be necessary now, try again

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +201,59 @@ def __get_rate(self, value):
...
 
 
+class DeviceUpgradeOperationInline(UpgradeOperationInline):
+
+    verbose_name = _('In progress upgrades')
+    verbose_name_plural = verbose_name
+
+    def get_queryset(self, request):
+        qs = super().get_queryset(request)
+        return qs.filter(status='in-progress')
```
nemesisdesign, line 247: I thought we said to show the last, so that it doesn't disappear when it's finished.
Ensure it's ordered by creation date (most recent first).

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +201,59 @@ def __get_rate(self, value):
...
     readonly_fields = ['installed', 'modified']
     verbose_name = _('Device Firmware')
     verbose_name_plural = verbose_name
     extra = 0
     multitenant_shared_relations = ['device']
 
 
+class DeviceUpgradeOperationInline(UpgradeOperationInline):
+
+    verbose_name = _('In progress upgrades')
```
nemesisdesign, line 242: Upgrade Operations

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +201,59 @@ def __get_rate(self, value):
...
+        qs = super().get_queryset(request)
+        return qs.filter(status='in-progress')
+
+
 def device_admin_get_inlines(self, request, obj):
-    inlines = self.inlines
+    # copy the list to avoid modifying the original data structure
+    inlines = list(self.inlines)
     if obj:
+        if obj.upgradeoperation_set.filter(status='in-progress').exists():
```
nemesisdesign, line 254: same here, we don't want that it disappears when the upgrade is complete (you still want to look to see how it went, if there were any errors or retries).

openwisp_firmware_upgrader/hardware.py
```diff
@@ -16,6 +16,10 @@
 OPENWRT_FIRMWARE_IMAGE_MAP.update(
     OrderedDict(
         (
+            (
+                'ramips-mt76x8-gl-mt300n-v2-squashfs-sysupgrade.bin',
+                {'label': 'GL.iNet GL-MT300N-V2', 'boards': ('GL-MT300N-V2',)},
+            ),
```
nemesisdesign, line 22: when you need to add this kind of stuff, you can just push directly to master, you should have the required permissions if I'm not mistaken

openwisp_firmware_upgrader/admin.py
nemesisdesign: Changed my mind, "Recent Upgrades" should be better

openwisp_firmware_upgrader/admin.py
nemesisdesign: Update: I think we can make it show only recent upgrades, so upgrades that have been created within something like a week

openwisp_firmware_upgrader/admin.py
nemesisdesign: I think you can do something like:

```python
DeviceUpgradeOperationInline(UpgradeOperation, admin.site).get_queryset().exists()
```

To avoid duplicating the queryset.

PabloCastellano: good trick

PabloCastellano:
For some reason the tests pass in my machine but fail in Travis :?

openwisp_firmware_upgrader/admin.py
```diff
@@ -1,23 +1,27 @@
...
 from django.utils.safestring import mark_safe
+from django.utils.timezone import localtime
 from django.utils.translation import ugettext_lazy as _
 from reversion.admin import VersionAdmin
 
 from openwisp_controller.config.admin import DeviceAdmin
 from openwisp_users.multitenancy import MultitenantAdminMixin
 from openwisp_utils.admin import ReadOnlyAdmin, TimeReadonlyAdminMixin
 
+# from .hardware import FIRMWARE_IMAGE_MAP
```
nemesisdesign, line 19: :point_up:

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +204,65 @@ def __get_rate(self, value):
...
 
 
+class DeviceFirmwareForm(forms.ModelForm):
+    class Meta:
+        model = load_model('DeviceFirmware')
+        fields = '__all__'
+
+    def _get_compatible_boards(self, instance):
+        # instance.device.model in instance.image.boards
+        # boards = FIRMWARE_IMAGE_MAP[instance.image.type]['boards']
```
nemesisdesign, line 214: :point_up:

setup.py
```diff
@@ -35,9 +35,10 @@
     zip_safe=False,
     install_requires=[
         # TODO: uncomment when they're released
-        # 'openwisp-utils[rest]~=0.5',
         # 'openwisp-controller~=0.7',
+        'openwisp-utils[rest]~=0.5',
         'django-private-storage~=2.2',
+        'python-dateutil~=2.8',
```
nemesisdesign, line 41: what is this needed for?

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +204,65 @@ def __get_rate(self, value):
...
+        device_model = instance.device.model
+        qs = load_model('FirmwareImage').objects.filter(
+            build__category__organization=instance.device.organization
+        )  # Or filter by type=...
+        list_of_ids = [
+            fw_image.pk for fw_image in qs if device_model in fw_image.boards
+        ]
```
nemesisdesign, line 221: take into account the case in which device_model is empty

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +204,65 @@ def __get_rate(self, value):
     aborted_rate.short_description = _('abortion rate')
 
 
+class DeviceFirmwareForm(forms.ModelForm):
+    class Meta:
+        model = load_model('DeviceFirmware')
+        fields = '__all__'
+
+    def _get_compatible_boards(self, instance):
```
nemesisdesign, line 212: I think this is not working, I see all the images, of different organizations, are there tests for this?

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,22 +204,65 @@ def __get_rate(self, value):
...
+        # instance.device.model in instance.image.boards
+        # boards = FIRMWARE_IMAGE_MAP[instance.image.type]['boards']
+        device_model = instance.device.model
+        qs = load_model('FirmwareImage').objects.filter(
+            build__category__organization=instance.device.organization
+        )  # Or filter by type=...
+        list_of_ids = [
```
nemesisdesign, line 219: I would also take in consideration the category.
Something like the following should work very well:

if the DeviceFirmware instance already exists, select images of the same category used by the current firmware, this will make upgrading very easy.
when the DeviceFirmware instance does not exist, show all categories.

PR #84, opened 2020-06-21
[api] Use uuid path converter in URLs

nemesisdesign:


PabloCastellano:
@nemesisdesign Any idea why travis fails?

PR #91, opened 2020-07-02
[private-storage] limit access to organization admin

### LINKED ISSUE #90: [fw-upgrader] Have private storage check can_access
This piece of code: https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/private_storage/views.py

Should ensure only users which are flagged as is_admin=True on the OrganizationUser of the related org can perform that action.

atb00ker:
Closes #90

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_api.py
+++ b/openwisp_firmware_upgrader/tests/test_api.py
@@ -33,6 +33,11 @@ class TestAPIUpgraderMixin(TestMultitenantAdminMixin, TestUpgraderMixin):
         self.operator = self._create_operator(organizations=[self.org])
         self._login()
 
+    def _make_operator_org_manager(self):
+        orgrelation = OrganizationUser.objects.get(user=self.operator)
+        orgrelation.is_admin = True
+        orgrelation.save()
+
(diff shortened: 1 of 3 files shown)


openwisp_firmware_upgrader/tests/test_api.py
```diff
@@ -757,9 +762,17 @@ def test_firmware_download(self):
...
         url = reverse('upgrader:api_firmware_download', args=[image.build.pk, image.pk])
-        with self.assertNumQueries(6):
-            r = self.client.get(url)
-        self.assertEqual(r.getvalue(), content)
+        with self.subTest("Test as operator"):
+            self._make_operator_org_manager()
+            with self.assertNumQueries(6):
+                response = self.client.get(url)
+            self.assertEqual(response.getvalue(), content)
+        with self.subTest("Test as superuser"):
```
atb00ker, line 770: Superuser is saved 3 queries tat happen in `user.is_manager()` usage.

nemesisdesign: probably because `self.object.build.category.organization` yuck

PR #96, opened 2020-07-09
feat: automatically create DeviceFirmware for Device instances #87

### LINKED ISSUE #87: [feature] Automatically create DeviceFirmware instances
When a device registers it will have no `DeviceFirmware` instance associated, which makes it cumbersome to upgrade (because it must be created from scratch and some users may not know what to do, while if it's already there, changing it is very easy).

Here's a simple way we could create it automatically.

All devices using the same firmware build will have the same content in the "operating system" field of the device.

We can add a new field to `Build`, something like `os_identifier`, where users have to place what comes up after devices register (users have to find out by extracting the firmware image with `binwalk` or by flashing a device).

When a device registers (`post_save` signal of `Device` with `created` argument `True`), we can check whether there's any Build for the same organization which has that os identifier and if there's a firmware image compatible with the model of the device, if there is, we can create the `DeviceFirmware` intance (making sure to not try to trigger an upgrade by using save(upgrade=False) when saving the FirmwareImage instance).

The opeartion should skip if the `os` and `model` fields are empty.

The receiver function should be connected in the [ready() method of the AppConfig, like we do in openwisp-monitoring](https://github.com/openwisp/openwisp-monitoring/blob/master/openwisp_monitoring/device/apps.py#L27-L53).

PabloCastellano:
Closes #87


openwisp_firmware_upgrader/base/models.py
```diff
@@ -266,6 +296,32 @@ def create_upgrade_operation(self, batch):
...
 
+    @classmethod
+    def auto_add_device_firmware_to_device(cls, instance, created, **kwargs):
+        # Automatically associate DeviceFirmware to the registered Device
+        if created:
+            if not instance.device.os or not instance.device.model:
+                return
+            if instance.device.model not in REVERSE_FIRMWARE_IMAGE_MAP:
+                return
+
```
nemesisdesign, line 307: if many devices exist (eg: hundreds) adding a connection with `auto_add` set to `True` is a bit slow already right now without this receiver, after adding them we risk to make it a bit too slow because the queries we'll be doing here are intensive, which may cause a timeout in the HTTP request.

Can you create a celery task for this?

After `if instance.device.model not in REVERSE_FIRMWARE_IMAGE_MAP:` we can then pass the execution to a background task:

```python
transaction.on_commit(lambda: create_device_firmware.delay(operation.pk))
```

This operation does not need to be synchronous so this would work well.

openwisp_firmware_upgrader/base/models.py
```diff
@@ -266,6 +296,32 @@ def create_upgrade_operation(self, batch):
...
+            DeviceFirmware = load_model('DeviceFirmware')
+            FirmwareImage = load_model('FirmwareImage')
+
+            try:
+                image = FirmwareImage.objects.get(
+                    build__category__organization_id=instance.device.organization_id,
+                    build__os_identifier=instance.device.os,
+                    type=REVERSE_FIRMWARE_IMAGE_MAP[instance.device.model],
+                )
+                device_fw = DeviceFirmware(
```
nemesisdesign, line 317: I think we can create a class method for this operation:

`DeviceFirmware.create_for_device(device)`, this method can execute this code.

I would also check if the `DeviceFirmware` already exists. It should not happen, but we never know.. and I get these kind of weird corner cases from time to time so now I want to start to avoid them to present themselves in the first place instead of having to debug them later on which is harder because we'll be busy doing other things.

So this method, before trying to create a new `DeviceFirmware`, we could do:

```python
qs = DeviceFirmware.objects.filter(device=dev)
if qs.exists():
    return qs.get()
```

We can then call this method in the celery task and in the migration.

openwisp_firmware_upgrader/migrations/__init__.py
```diff
@@ -53,3 +58,8 @@ def create_permissions_for_default_groups(apps, schema_editor, app_label):
                 content_type__app_label=app_label,
             )
             admin.permissions.add(permission.pk)
+
+
+def create_device_firmware_for_connections(apps, schema_editor, app_label):
+    for connection in DeviceConnection.objects.all():
+        connection.save()
```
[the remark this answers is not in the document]
nemesisdesign: here we could call:
```python
DeviceFirmware.create_for_device(device_connection.device)
```

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -124,6 +146,61 @@ def test_device_fw_no_connection(self):
...
+        org = self._get_org()
+        category = self._get_category(organization=org)
+        build1 = self._create_build(
+            category=category, version='0.1', os_identifier=os_identifier
+        )
+        image1a = self._create_firmware_image(
+            build=build1, type=REVERSE_FIRMWARE_IMAGE_MAP['YunCore XD3200']
+        )
+
+        with self.subTest('Device doesn\'t define os'):
```
nemesisdesign, line 160: in these cases it's ok to use the double quote (`"`) if you need

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -124,6 +146,61 @@ def test_device_fw_no_connection(self):
...
+                model=image1a.boards[0], os=os_identifier
+            )
+            self.assertEqual(DeviceConnection.objects.count(), 1)
+            self.assertEqual(Device.objects.count(), 1)
+            self.assertEqual(DeviceFirmware.objects.count(), 0)
+
+    def test_device_fw_created_on_device_connection_save(self):
+        self.assertEqual(DeviceFirmware.objects.count(), 0)
+        self.assertEqual(Device.objects.count(), 0)
+        os_identifier = 'Linux'
```
nemesisdesign, line 190: let's use a more realistic os identifier in our tests, just to make it more familiar to future devs, eg: `OpenWrt 19.07-SNAPSHOT r11061-6ffd4d8a4d`.

openwisp_firmware_upgrader/apps.py
```diff
@@ -33,3 +36,12 @@ def add_default_menu_items(self):
...
             current_menu += items
+
+    def connect_device_signals(self):
+        DeviceConnection = load_model('connection', 'DeviceConnection')
+        DeviceFirmware = load_model('firmware_upgrader', 'DeviceFirmware')
+        post_save.connect(
+            DeviceFirmware.auto_add_device_firmware_to_device,
+            sender=DeviceConnection,
+            dispatch_uid='connection.auto_add_device_firmware',
+        )
```
nemesisdesign, line 47: I thought about another use case that I'm sure it will happen for reasons I won't explain here for brevity.

It will happen that default credentials with `auto_add` are already defined, hence a `DeviceConnection` will be created before firmware images are uploaded into the syste, which will cause devices to be orphans of `DeviceFirmware` again.

To avoid this it should be enough to add another receiver to the `post_save` of `FirmwareImage`, something like this:
```python
def receiver_xyz(instance, created):
    if created:
        transaction.on_commit(lambda: create_all_device_firmwares.delay(instance.pk))
```

The task would be something like:
```python
@shared_task
create_all_device_firmwares(firmware_image_pk):
    queryset = Device.objects.filter(os=fw_image.build.os)
    if not queryset.exists():
        return
    fw_image = FirmwareImage.objects.get(pk=firmware_image_pk).select_related('build')
    for device in queryset:
        DeviceFirmware.create_for_device(device, firmware_image)
```

In this example I added the ``firmware_image`` argument, since we know it beforehand it would be good to avoid querying the DB again.

PabloCastellano:
@nemesisdesign Thanks for the complete review. Now for some reason the celery task is not executed and the test fail. Any idea?

PR #104, opened 2020-09-24
[change] Added notification module (from openwisp-controller)

nemesisdesign:
Added notification module (from openwisp-controller).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,8 +61,6 @@ Follow the setup instructions of `openwisp-controller
 
     INSTALLED_APPS = [
         # django apps
-        # openwisp2 admin theme (must be loaded here)
-        'openwisp_utils.admin_theme',
         # all-auth
         'django.contrib.sites',
         'allauth',
@@ -70,12 +68,15 @@ Follow the setup instructions of `openwisp-controller
         'django_extensions',
         'private_storage',
         # openwisp2 modules
-        'openwisp_users',
         'openwisp_controller.pki',
         'openwisp_controller.config',
         'openwisp_controller.connection',
         'openwisp_controller.geo',
         'openwisp_firmware_upgrader',
+        'openwisp_users',
+        'openwisp_notifications',
+        # openwisp2 admin theme (must be loaded here)
+        'openwisp_utils.admin_theme',
         # admin
         'django.contrib.admin',
         'django.forms',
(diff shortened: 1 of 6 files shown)


nemesisdesign:
@pandafy do you have any idea of why the notifications queries get executed in the firmware-upgrader API?

A quick investigation is leading me to believe that those queries are triggered by the signal receivers but I would like to hear your opinion on the matter. We need to keep the amount of queries under control because as we're adding more features the system risks to become too slow and won't scale with high load.

See: https://travis-ci.org/github/openwisp/openwisp-firmware-upgrader/jobs/730092341#L1153-L1256

pandafy:
@nemesisdesign I didn't get time to investigate it further today, will try to do it tomorrow. 

Can you please add `INTERNAL_IPS = ['127.0.0.1']` in `settings.py`? Otherwise websocket will try to connect with `wss` protocol on development server.

pandafy:
Thanks for your patience Federico! 

I looked up the queries which are generated after introducing openwisp_notification through openwisp_controller. I don't think there is anything to worry about those database queries as they don't affect system responsiveness directly. I have described the reasons for saying so below.

```
SELECT COUNT(*) AS "__count" FROM "openwisp_notifications_notification" WHERE ("openwisp_notifications_notification"."recipient_id" = 'd07bc82efc554d8cad4f17acf5407920' AND "openwisp_notifications_notification"."unread")
```
This query is made to get the total number of unread notifications to show on Django admin. This value is generally served from cache therefore database lookups are occasional only when the value is getting updated. 

```
SELECT "openwisp_notifications_notification"."id", "openwisp_notifications_notification"."level", "openwisp_notifications_notification"."recipient_id", "openwisp_notifications_notification"."unread", "openwisp_notifications_notification"."actor_content_type_id", "openwisp_notifications_notification"."actor_object_id", "openwisp_notifications_notification"."verb", "openwisp_notifications_notification"."description", "openwisp_notifications_notification"."target_content_type_id", "openwisp_notifications_notification"."target_object_id", "openwisp_notifications_notification"."action_object_content_type_id", "openwisp_notifications_notification"."action_object_object_id", "openwisp_notifications_notification"."timestamp", "openwisp_notifications_notification"."public", "openwisp_notifications_notification"."deleted", "openwisp_notifications_notification"."emailed", "openwisp_notifications_notification"."data", "openwisp_notifications_notification"."type" FROM "openwisp_notifications_notification" WHERE (("openwisp_notifications_notification"."actor_content_type_id" = 27 OR "openwisp_notifications_notification"."action_object_content_type_id" = 27 OR "openwisp_notifications_notification"."target_content_type_id" = 27) AND ("openwisp_notifications_notification"."actor_object_id" = '3104cb58-1c83-4df6-814f-5e71c6d20183' OR "openwisp_notifications_notification"."action_object_object_id" = '3104cb58-1c83-4df6-814f-5e71c6d20183' OR "openwisp_notifications_notification"."target_object_id" = '3104cb58-1c83-4df6-814f-5e71c6d20183'))
7. SELECT "openwisp_notifications_ignoreobjectnotification"."id", "openwisp_notifications_ignoreobjectnotification"."user_id", "openwisp_notifications_ignoreobjectnotification"."object_content_type_id", "openwisp_notifications_ignoreobjectnotification"."object_id", "openwisp_notifications_ignoreobjectnotification"."valid_till" FROM "openwisp_notifications_ignoreobjectnotification" WHERE ("openwisp_notifications_ignoreobjectnotification"."object_content_type_id" = 27 AND "openwisp_notifications_ignoreobjectnotification"."object_id" = '3104cb58-1c83-4df6-814f-5e71c6d20183')
```

You are right, [there is a receiver for pre-delete signal in openwisp_notifications](https://github.com/openwisp/openwisp-notifications/blob/master/openwisp_notifications/handlers.py#L232) which deletes objects of `Notification` and `IgnoreObjectNotification` model. Since, these models have GenericForeignKey, the database cascade does not work and we need to manually delete such objects which have become obsolete. 

These queries to are made asynchronously through celery tasks, therefore does not hinder system responsive directly. 

I wonder why they show up as `SELECT` queries in the traceback.

nemesisdesign:
Thanks for double checking @pandafy :pray:

Issue #35, opened 2020-04-20
[fw-upgrade] Flag logging strings as translatable


nemesisdesign:
The strings used to store log messages in the upgrader class should be translatable, because these lines are then shown to the user.
Look for `self.log(`: https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/upgraders/openwrt.py

See the django docs regarding translatable strings: https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#standard-translation

morrme:
Hi! Is this still a reserved issue? Either way, do you require the self.log() be replaced with something like gettext() as in the referenced Django docs?

nikhanal:
Can I work on this issue?
Can you please guide me through because I am new to open source projects

nemesisdesign:
@morrme yes, the abbreviation `_(message)` to be more precise.
@nikhanal all the info is described in the issue description, if you have more specific questions, do not hesitate to ask.

nikhanal:
should it look like self.log._(message)? or self.log(message) should be completely replaced by _(message)?

atb00ker:
@nikhanal , I think it should be `self.log(_(message))`; it's basic python, think about it, it would make sense. :smile:
Let me know if you still need help with it. :smile:

PR #106, opened 2020-10-14
[fix] Made celery tasks resilient to failure #98

### LINKED ISSUE #98: [celery] Make tasks resilient to failures
In some rare cases, the tasks `upgrade_firmware` and `batch_upgrade_operation` may try to execute tasks on objects which have already been deleted.

We need to catch the DoesNotExist exception, log a warning but avoid the exception to be raised, similarly to what it's being done here: https://github.com/openwisp/openwisp-monitoring/pull/192/files

darecoder:
- Added exception handling for "upgrade_firmware"
  and "batch_upgrade_operation" tasks
- Fixed minor typographical errors in test_tasks
- Made log warning translatable

Fixes #98

### DIFF
--- a/openwisp_firmware_upgrader/tasks.py
+++ b/openwisp_firmware_upgrader/tasks.py
@@ -2,6 +2,7 @@ import logging
 
 from celery import shared_task
 from celery.exceptions import SoftTimeLimitExceeded
+from django.core.exceptions import ObjectDoesNotExist
 from django.utils.translation import gettext_lazy as _
 
 from openwisp_controller.config.models import Device
@@ -17,7 +18,7 @@ logger = logging.getLogger(__name__)
     bind=True,
     autoretry_for=(RecoverableFailure,),
     soft_time_limit=app_settings.TASK_TIMEOUT,
-    **app_settings.RETRY_OPTIONS
+    **app_settings.RETRY_OPTIONS,
 )
 def upgrade_firmware(self, operation_id):
     """
@@ -30,8 +31,12 @@ def upgrade_firmware(self, operation_id):
         operation.upgrade(recoverable=recoverable)
     except SoftTimeLimitExceeded:
         operation.status = 'failed'
-        operation.log_line(_('Operation timed out.'))
+        operation.log_line('Operation timed out.')
         logger.warn('SoftTimeLimitExceeded raised in upgrade_firmware task')
+    except ObjectDoesNotExist:
+        logger.warn(
+            f'The UpgradeOperation object with id {operation_id} has been deleted'
+        )
 
 
 @shared_task(bind=True, soft_time_limit=app_settings.TASK_TIMEOUT)
@@ -47,6 +52,10 @@ def batch_upgrade_operation(self, batch_id, firmwareless):
         batch_operation.status = 'failed'
         batch_operation.save()
         logger.warn('SoftTimeLimitExceeded raised in batch_upgrade_operation task')
+    except ObjectDoesNotExist:
+        logger.warn(
+            f'The BatchUpgradeOperation object with id {batch_id} has been deleted'
+        )
 
 
 @shared_task(bind=True)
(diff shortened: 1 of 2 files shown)


openwisp_firmware_upgrader/tests/test_tasks.py
```diff
@@ -42,3 +42,27 @@ def test_batch_upgrade_timeout(self, *args):
...
+            f'The UpgradeOperation object with id {upgrade_op.id} has been deleted'
+        )
+
+    @mock.patch(_mock_upgrade, return_value=True)
+    @mock.patch(_mock_connect, return_value=True)
+    @mock.patch('logging.Logger.warning')
+    def test_batch_upgrade_operation_resilience(self, mocked_logger, *args):
+        env = self._create_upgrade_env()
+        batch = BatchUpgradeOperation.objects.create(build=env['build2'])
+        batch.delete()
```
nemesisdesign, line 64: there's no need to actually create and delete the object, you can simply instantiate it to get the uuid: `BatchUpgradeOperation().pk`, this makes the test faster to run, which is good because `TransactionTestCase` is slow.

openwisp_firmware_upgrader/tests/test_tasks.py
```diff
@@ -42,3 +42,27 @@ def test_batch_upgrade_timeout(self, *args):
...
         batch = BatchUpgradeOperation.objects.first()
         self.assertEqual(batch.status, 'failed')
+
+    @mock.patch(_mock_upgrade, return_value=True)
+    @mock.patch(_mock_connect, return_value=True)
+    @mock.patch('logging.Logger.warning')
+    def test_upgrade_firmware_resilience(self, mocked_logger, *args):
+        device_fw = self._create_device_firmware(upgrade=True)
+        upgrade_op = device_fw.image.upgradeoperation_set.first()
+        upgrade_op.delete()
```
nemesisdesign, line 52: same here, you can trick it with `UpgradeOperation().pk`

nemesisdesign (changes requested):
Thanks a lot for contributing @darecoder!

It looks good. The only thing that does not convince me is making the log messages translatable, mainly because we don't do this in other modules, nor we plan on translating the log messages, since we take for granted that those who will read the log messages understand coding and understand English pretty well. So for a matter of consistency can you use plain strings instead of translatable strings?

darecoder:
Thanks for the review @nemesisdesign. Actually, I saw that there is already an [issue](https://github.com/openwisp/openwisp-firmware-upgrader/issues/35) to make logging strings translatable, that's why I added them here.

nemesisdesign:
> Thanks for the review @nemesisdesign. Actually, I saw that there is already an [issue](https://github.com/openwisp/openwisp-firmware-upgrader/issues/35) to make logging strings translatable, that's why I added them here.

Thanks for bringing this up, I forgot about that issue. Let me close it, it's really not a useful use of our time.

nemesisdesign (commented):
> Thanks for the review @nemesisdesign. Actually, I saw that there is already an issue to make logging strings translatable, that's why I added them here.
> 
> Thanks for bringing this up, I forgot about that issue. Let me close it, it's really not a useful use of our time.

I correct myself: that issue is about the `log` method of the upgrader, which saves messages which are shown to users.

The python logger used to log messages for developers and system administrators who need to debug the application is one thing (what you did here), that does not need translation.

The log() method in the upgrader classes save output which is shown to the user, that needs to be translated.

I hope I made the issue clearer.

Thanks

openwisp_firmware_upgrader/tasks.py
```diff
@@ -30,8 +31,12 @@ def upgrade_firmware(self, operation_id):
         operation.upgrade(recoverable=recoverable)
     except SoftTimeLimitExceeded:
         operation.status = 'failed'
-        operation.log_line(_('Operation timed out.'))
+        operation.log_line('Operation timed out.')
```
nemesisdesign, line 34: this is shown to the user as well, it shouldn't be changed @darecoder

nemesisdesign: I fixed it, will squash my changes. Thanks.

nemesisdesign (approved):

PR #114, opened 2020-12-25
[qa] Add admin checks for details field #61

### LINKED ISSUE #61: [qa] Write admin tests which check the presence of details field in admin UI
The sample app models have a details field which is shown in the admin:

https://github.com/openwisp/openwisp-firmware-upgrader/blob/1136746050f98851d0dc0fb75ec37cb1b02bd10b/tests/openwisp2/sample_firmware_upgrader/models.py#L12-L16

But we don't have automated tests for this.

If this feature breaks, we won't know from tests.

Write additional tests in `tests/openwisp2/sample_firmware_upgrader/tests.py` (TestAdmin) to check this.

One test for each admin class being used.

Purhan:
Closes #61

### DIFF
--- a/tests/openwisp2/sample_firmware_upgrader/admin.py
+++ b/tests/openwisp2/sample_firmware_upgrader/admin.py
@@ -2,4 +2,8 @@ from openwisp_firmware_upgrader.admin import (  # noqa
     BatchUpgradeOperationAdmin,
     BuildAdmin,
     CategoryAdmin,
+    UpgradeOperationForm,
 )
+
+BatchUpgradeOperationAdmin.fields.append('details')
+UpgradeOperationForm.Meta.fields.append('details')
--- a/tests/openwisp2/sample_firmware_upgrader/tests.py
+++ b/tests/openwisp2/sample_firmware_upgrader/tests.py
@@ -81,6 +81,30 @@ class TestAdmin(BaseTestAdmin):
             'value="sample device_fw details" class="vTextField"',
         )
 
+    def test_batch_upgrade_operation_details(self):
+        self._login()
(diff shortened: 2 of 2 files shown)


Purhan (commented):
How can I test `BatchUpgradeOperation` and `UpgradeOperation`? No endpoints show details in the admin interface for these 2 models. Please help.

Purhan:
> How can I test `BatchUpgradeOperation` and `UpgradeOperation`? No endpoints show details in the admin interface for these 2 models. Please help.

The lines below are the reason why. I could overwrite this in the `sample_app/admin.py` to include the details field. Is that a good option?
https://github.com/openwisp/openwisp-firmware-upgrader/blob/d5d82f6442875737f9da6834d9bc558c7af0f7c3/openwisp_firmware_upgrader/admin.py#L194-L203

pandafy:
I think we can also can monkey patch the respective admin class as show in the [documentation](https://github.com/openwisp/openwisp-firmware-upgrader#1-monkey-patching). I believe that should be sufficient for testing. What do you think?

Purhan:
> I think we can also can monkey patch the respective admin class as show in the [documentation](https://github.com/openwisp/openwisp-firmware-upgrader#1-monkey-patching). I believe that should be sufficient for testing. What do you think?

Yes, that's exactly what I mean by overwriting in `sample_app/admin.py`, had to use the correct terminology :smile:

Purhan:
@pandafy Thanks for helping out! I'm marking this as ready for review now. Please take a look/

nemesisdesign (approved):
Great work @Purhan! Thanks @pandafy for reviewing. :+1:

PR #117, opened 2021-01-21
[fw-upgrader] Avoid deletion of UpgradeOperation when FirmwareImage is deleted

### LINKED ISSUE #66: [fw-upgrader] Do not delete upgrade operations when FirmwareImage is deleted
It should be possible to delete firmware images to free space but history shouldn't be lost.

1. Add a failing test for this case
2. Apply fix, setting on_delete=models.SET_NULL to UpgradeOperation.firmware_image should be enough

R9295:
closes #66


openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -249,6 +249,14 @@ def test_create_for_device_validation_error(self):
         result = DeviceFirmware.create_for_device(device_fw.device)
         self.assertIsNone(result)
 
+    def test_upgrade_operation_retention_on_image_delete(self):
+        device_fw = self._create_device_firmware()
+        uo = UpgradeOperation.objects.create(
+            device=device_fw.device, image=device_fw.image
+        )
+        FirmwareImage.objects.get(pk=device_fw.image.pk).delete()
+        self.assertEqual(UpgradeOperation.objects.get(pk=uo.pk).image, None)
```
pandafy, line 258: The goal of this test is to check whether objects of `UpgradeOperation` are available even after the corresponding `FirmwareImage` is deleted. 

The current assertion makes sure such object exists since `UpgradeOperation.objects.get(pk=uo.pk)` is not raising `DoesNotExist` exception. But, it is not evident from the code at first glance. 

I suggest to first ensure that `UpgradeOperation` object exists in a assertion and then check it's image attribute is set to `None` in a separate assertion.

The above suggestion will make code more readable.

tests/openwisp2/sample_firmware_upgrader/migrations/0007_create_device_firmware.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.0.8 on 2020-07-07 17:00
```
pandafy, line 1: I am not sure whether this migration is really needed. From the code, I understand that `create_device_firmware_for_connections` will try to create a `DeviceFirmware` instance for existing devices. From the scope of sample app, I don' think it is necessary. 

@nemesisdesign what do you think about this?

pandafy (changes requested):
Good job @R9295!

Kindly make sure that the commit message follows [OpenWISP commit guidelines](https://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines). From next time, open a PR from branch other than `master` as suggested in [OpenWISP's branch naming guidelines](https://openwisp.io/docs/developer/contributing.html#branch-naming-guidelines).


P.S.: I was not able to test Batch Upgrade Operation. Will report if I found any issues.

tests/openwisp2/sample_firmware_upgrader/migrations/0008_nullable_upgrade_operation_image.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.5 on 2021-01-21 16:57
```
pandafy, line 1: @nemesisdesign I saw the number of migrations in sample app and many are just tweaking some attributes. Do you think we should squash such migration files into one or just update the initial migrations for sanity(just for sample app)?

nemesisdesign: yes in general I ask to re-generate the migrations but in this case since the migrations are many we could do it in a separate issue:
https://github.com/openwisp/openwisp-firmware-upgrader/issues/119

tests/openwisp2/sample_firmware_upgrader/migrations/0007_create_device_firmware.py
nemesisdesign: @pandafy you're correct, although it can be useful for developers wanting to extend fw-upgrader and looking for a reference implementation, so we can leave this data migration.
@R9295 why did you add this migration? Is it because you wanted the two apps to have the same number of migrations?
We can keep it for the reason expressed above.

nemesisdesign (approved):
Thanks @R9295 and @pandafy for your help in reviewing.

I tested it and found no issues, the code also looks good.

Let's do the cleanup of the migration files in https://github.com/openwisp/openwisp-firmware-upgrader/issues/119.

PR #122, opened 2021-01-26
[fix] Replace Device model references with swapper calls

okraits:


openwisp_firmware_upgrader/tasks.py
```diff
@@ -66,6 +65,7 @@ def create_device_firmware(self, device_id):
     if qs.exists():
         return
 
+    Device = swapper.load_model('config', 'Device')
```
nemesisdesign, line 68: you should be able to place this at the top of the file and just once, the same suggestion is valid for the other occurrences, please try and let me know

nemesisdesign (changes requested):

openwisp_firmware_upgrader/tasks.py
okraits: I thought about that too, but then, when running `run-qa-checks`, I get `django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.`.

nemesisdesign: you're right, in tasks we need to do this inside the functions, although in the next version of swapper we'll be able to use `require_ready=False`: https://github.com/openwisp/django-swappable-models/commit/1ac54fc06e91d41acb0b4a52f877e711b4c6a6e4

nemesisdesign (approved):

PR #124, opened 2021-03-12
[cleanup] Re-generated sample app migrations #119

codesankalp:
Regenerated all migrations except data migration.

Closes #119

PR #126, opened 2021-03-16
[fix] Removed apt-get upgrade as it cause ci issues

devkapilbansal:


nemesisdesign (commented):
@devkapilbansal should have been solved by #124, right?

devkapilbansal:
My bad, I didn't noticed that this has been fixed in another PR.  :grimacing: 
Closing this PR as it is no longer needed

PR #131, opened 2021-03-25
Removed duplicate setup instructions

atb00ker:
The same instructions are there on openwisp-controller as well!


nemesisdesign (approved):
Thanks :+1: (i'll fix the commit message).

PR #139, opened 2021-04-18
[settings]  Made FIRMWARE_API_BASEURL functionationlity generic

atb00ker:


openwisp_firmware_upgrader/private_storage/views.py
```diff
@@ -8,7 +8,7 @@ class FirmwareImageDownloadView(PrivateStorageDetailView):
     model_file_field = 'file'
 
     slug_field = 'file'
-    slug_url_kwarg = 'path'
+    slug_url_kwarg = 'imagefile'
```
nemesisdesign, line 11: why are you changing this and are you sure this is backward compatible?

atb00ker: I changed it because `path` too vague and at rest of the places the name is, example `csvfile`. 
I don't think it should have any compatibility issue, it's not exposed to anything in a stateful way, just the variable name that stores the incoming file! :smile:

nemesisdesign (approved):
Did a quick test to verify images uploaded before keep being downloadable, let's proceed!

Thanks.

PR #149, opened 2021-08-09
[fix] DeviceAdmin.conditional_inlines were not added to edit form

pniaps:
openwisp-firmware-upgrader is overriding DeviceAdmin.get_inlines, so inlines added to DeviceAdmin.conditional_inlines are not shown.

This commit calls the old DeviceAdmin.get_inlines to get the inlines and the conditional inlines.

The problem was detected because SSH commands was not shown when using openwisp-firmware-upgrader

### DIFF
--- a/openwisp_firmware_upgrader/admin.py
+++ b/openwisp_firmware_upgrader/admin.py
@@ -325,9 +325,13 @@ class DeviceUpgradeOperationInline(UpgradeOperationInline):
         return qs
 
 
+# TODO:
+# DeviceAdmin.inlines += [DeviceFirmwareInline]
+# DeviceAdmin.conditional_inlines += [DeviceUpgradeOperationInline]
+old_device_admin_get_inlines = DeviceAdmin.get_inlines
 def device_admin_get_inlines(self, request, obj):
     # copy the list to avoid modifying the original data structure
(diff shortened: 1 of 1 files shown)


openwisp_firmware_upgrader/admin.py
```diff
@@ -325,9 +325,15 @@ def get_queryset(self, request, select_related=True):
         return qs
 
 
+# TODO:
+# DeviceAdmin.inlines += [DeviceFirmwareInline]
+# DeviceAdmin.conditional_inlines += [DeviceUpgradeOperationInline]
```
pandafy, line 330: I think you can remove this.

nemesisdesign (commented):
I think this was already fixed in the `dev` branch:
https://github.com/openwisp/openwisp-firmware-upgrader/blob/dev/openwisp_firmware_upgrader/admin.py#L337

Can you please double check if that branch resolves your issue?

pniaps:
Yes, tested.

Thanks.

openwisp_firmware_upgrader/admin.py
pniaps: I was there to remember that solution is better, but I have doubs

PR #150, opened 2021-08-11
[refactor] Switch to new openwisp-users API mixins & classes #142

### LINKED ISSUE #142: [change] API: use mixins from openwisp-users
We should change the code of this module to use the new API classes and mixins added to openwisp-users, which were based on the code from this module so should be backward compatible, however, I found out a few issues:

- some classes in openwisp-users were not taking into account the `organization_lookup` field, hence queries were failing to find the right organization field, I solved this in https://github.com/openwisp/openwisp-users/tree/improve-api-mixins, but tests are missing, we need at least one test in openwisp-users. Issue: https://github.com/openwisp/openwisp-users/issues/255.
- a lot of `assertNumQueries` tests are failing! It looks like the number of queries increased between 2 and 3 queries for each endpoint, we should understand what the additional queries are doing, if they're really needed or not, if the extra queries are generated by mistake or not and what we can do to fix it. Issue: https://github.com/openwisp/openwisp-users/issues/256

I started this work on this branch: https://github.com/openwisp/openwisp-firmware-upgrader/tree/openwisp-users-api-mixins, it needs this other branch from openwisp-users to work: https://github.com/openwisp/openwisp-users/tree/improve-api-mixins, I had to give up because it's taking longer than expected and I hope @ManishShah120 you can help me to finish this. 

PS: I think we should also change the definitions of the API views to use tuples instead of lists.

ManishShah120:
Fixes #142

### DIFF
--- a/openwisp_firmware_upgrader/api/views.py
+++ b/openwisp_firmware_upgrader/api/views.py
@@ -3,11 +3,12 @@ from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import filters, generics, pagination, serializers
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.exceptions import NotFound
-from rest_framework.permissions import DjangoModelPermissions
 from rest_framework.response import Response
 
 from openwisp_firmware_upgrader import private_storage
 from openwisp_users.api.authentication import BearerAuthentication
+from openwisp_users.api.mixins import FilterByOrganizationManaged
+from openwisp_users.api.permissions import DjangoModelPermissions, IsOrganizationManager
 
 from ..swapper import load_model
 from .serializers import (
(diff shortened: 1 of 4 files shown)


ManishShah120:
I have explained the reason behind the number of queries increasing here:-https://github.com/openwisp/openwisp-firmware-upgrader/issues/142#issuecomment-866947374

openwisp_firmware_upgrader/api/views.py
```diff
@@ -30,33 +31,32 @@ class ListViewPagination(pagination.PageNumberPagination):
...
     def get_queryset(self):
-        queryset = self.queryset.all()
-        if not self.request.user.is_superuser:
-            filter_key = f'{self.organization_field}__in'
-            user_orgs = self.request.user.organizations_dict.keys()
-            organization_filter = {filter_key: user_orgs}
-            queryset = queryset.filter(**organization_filter)
-        org = self.request.query_params.get('organization', None)
+        qs = super().get_queryset()
+        org_filtered = self.request.query_params.get('organization', None)
```
atb00ker, line 48: What about superusers, I don't see the code to give them access to everything!

ManishShah120: @atb00ker this portion of code is to filter the queryset based on `organization` params if provided in the request, the queyset objects will be taken care by `ProtectedAPIMixin`, and `FilterByOrganizationManaged`.

atb00ker: Read the code for the old version of this function, it returns `self.queryset.all()` for superusers, that's what I am talking about. We don't want any filtering for superusers, correct? 
They have global access to anything and everything.
There are no checks for superusers.

ManishShah120: @atb00ker if I can understand it correctly this portion of code is to filter the already received queryset based on the organization slug value, by passing a query param to the endpoint. if not passed this will returned whatever it gets from the class `FilterByOrganizzationManaged`.

PS: This portion of code was written by @nemesisdesign I have commited on top of it, by fixing the mixin classes and asseting the number of queries.

PR #152, opened 2021-08-18
[fix] Fixed button style in build page

niteshsinha17:
Fixed button style.
image

### DIFF
--- a/openwisp_firmware_upgrader/static/admin/css/firmware-upgrader.css
+++ b/openwisp_firmware_upgrader/static/admin/css/firmware-upgrader.css
@@ -1,11 +1,9 @@
 input.object-tools-button {
-    padding: 5px 15px;
-    background-color: #999;
-    font-weight: 400;
-    font-size: 13px;
+    padding: 0.4rem 0.8rem;
+    text-transform: uppercase;
     line-height: 20px;
+    display: inline-block;
+    font-size: 13px;
     margin: 0;
-    display: block;
-    float: left;
-    border-radius: 15px;
+    letter-spacing: 0.5px;
 }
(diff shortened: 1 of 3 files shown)


openwisp_firmware_upgrader/static/admin/css/firmware-upgrader.css
```diff
@@ -1,11 +1,9 @@
...
+    padding: 0.4rem 0.8rem;
+    text-transform: uppercase;
     line-height: 20px;
+    display: inline-block;
+    font-size: 13px;
     margin: 0;
-    display: block;
-    float: left;
-    border-radius: 15px;
+    letter-spacing: 0.5px;
```
nemesisdesign, line 8: is it possible to make sure the style is provided by openwisp-utils and remove any custom CSS here?
So that whenever we need to add a button like this one in any module we don't need to provide attitional CSS?

nemesisdesign (changes requested):

openwisp_firmware_upgrader/static/admin/css/firmware-upgrader.css
niteshsinha17: Actually here it is using ``input`` element that's why we need to add this css. Also, buttons present in ``object-tools`` have a similar style to the ``default button`` with difference in padding. they do not have any css class name but they styled relative to their position like 
```
.object-tools ul li a{
//
}
```

nemesisdesign: Maybe then this can be updated: https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/templates/admin/firmware_upgrader/change_form.html

niteshsinha17: Removed CSS file but now it depends on this https://github.com/openwisp/openwisp-utils/pull/246

nemesisdesign (approved):
Excellent, I am very happy we simplified the code here, this additional CSS should have never been added in the first place.

Issue #155, opened 2021-08-28
[feature] Add support sysupgrade options such as -n and --force


mips171:
OpenWISP firmware upgrader does not seem to support `sysupgrade -n` or `-F` upgrades.  This feature may be necessary if a _swconfig to to DSA_ conversion script cannot be implemented in OpenWrt because currently sysupgrade will fail with an image incompatibility error when upgrading.

PR #156, opened 2021-08-29
WIP/RFC: [feature] Support sysupgrade options --force and -n

mips171:
Closes https://github.com/openwisp/openwisp-firmware-upgrader/issues/155

This WIP/RFC adds support for specifying `sysupgrade` options  `--force` and `-n` no preserve settings.  `-n` and the default `-c` are mutually exclusive, so we only add the `-c` if we are not wiping the settings.

When I started this I felt like it should be part of an UpgradeOperation, not a FirmwareImage.  Would that make more sense?

I feel like the option to `-F` should be allowed, but we need to really stress to the user that unless they know what they are doing, they could end up with a potentially very far away brick.  So, I was thinking we either remove that altogether, or add an extra field, kind of like github's delete repo confirmation, that requires a string like `I want a brick` or `I understand the risks` or maybe the MAC address of the device to be included along with the boolean.  I suppose this could also be secured further via permissions.

firmware new options

sysupgrade command with options


openwisp_firmware_upgrader/base/models.py
```diff
@@ -254,6 +254,12 @@ class AbstractDeviceFirmware(TimeStampedEditableModel):
     )
     image = models.ForeignKey(get_model_name('FirmwareImage'), on_delete=models.CASCADE)
     installed = models.BooleanField(default=False)
+    no_preserve_settings = models.BooleanField(default=False)
```
nickberry17, line 257: I think we need a better way to display this to the user without including it as part of a FirmwareImage

nemesisdesign (commented):
Great points @nickberry17!

Definitely the fields make sense on the upgrade operations models and not there, I think it has to be achieved by changing different parts:

- a custom ModelForm with the additional fields, so that we don't have to add the fields on DeviceFirmware model
- pass the parameters in the internal methods starting from the `save()` method of DeviceFirmware:

https://github.com/openwisp/openwisp-firmware-upgrader/blob/c9d6dab745add5ce1d048acc3de6f45f39f21868/openwisp_firmware_upgrader/base/models.py

Basically, the custom ModelForm has additional parameters which are then passed to the `save()` method.
See:

https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.form
- https://docs.djangoproject.com/en/3.1/topics/forms/modelforms/

Let's also not forget that upgrade operations can be done also by creating a `BatchUpgradeOperation`

Something similar has to be implemented for the REST API too.

mips171:
> Great points @nickberry17!
> 
> Definitely the fields make sense on the upgrade operations models and not there, I think it has to be achieved by changing different parts:
> 
> * a custom ModelForm with the additional fields, so that we don't have to add the fields on DeviceFirmware model
> * pass the parameters in the internal methods starting from the `save()` method of DeviceFirmware:
> 
> https://github.com/openwisp/openwisp-firmware-upgrader/blob/c9d6dab745add5ce1d048acc3de6f45f39f21868/openwisp_firmware_upgrader/base/models.py
> 
> Basically, the custom ModelForm has additional parameters which are then passed to the `save()` method. See:
> 
> https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.form
> 
> * https://docs.djangoproject.com/en/3.1/topics/forms/modelforms/
> 
> Let's also not forget that upgrade operations can be done also by creating a `BatchUpgradeOperation`
> 
> Something similar has to be implemented for the REST API too.

Whats the LoE estimate for this one? Still would like to get this feature in

nemesisdesign:
@mips171 it's more complex than initially thought. Since it's possible to suppor custom firmwares or other systems, the options of the upgrader depend on the Upgrader class and we'll need to update the REST API too. It will take a few full days (4 to 5) to implement this consistently with the OpenWISP design.

PR #157, opened 2021-08-29
[models] Fixed 500 error on empty category #153

### LINKED ISSUE #153: Error 500 when you add a new build without selecting category
When you go to **Firmware Builds** > **Add Firmware Build**, fill **Version** and **OS identifier** and click **Save**, there is an error 500.

In debug, it says **Build has no category**.

If you don't fill **OS identifier**, the screen is fine, **Firmware category** has error **This field is required**

(i'm testing `dev` branch)

hanif-ali:
Closes #153 

Problem:
Trying to add a Firmware Build with an empty category but a non-empty OS identifier caused a 500 error.
The problem was in a uniqueness check that got executed before `category` field was validated. The code checked for unique (`category.organization`, `os`) for a build.
With an empty `os` field, this check was skipped (`if not self.os: return`) so the rest of the validation checks ran normally.

This change makes it so that the check is skipped when either `os` or `category` fields are empty. The empty field is caught in the rest of the validation checks and reported to the user. Test for the bug has been added.

### DIFF
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -113,9 +113,7 @@ class AbstractBuild(TimeStampedEditableModel):
             return
         if (
             load_model('Build')
-            .objects.filter(
-                category__organization=category.organization, os=self.os
-            )
+            .objects.filter(category__organization=category.organization, os=self.os)
             .exclude(pk=self.pk)
             .exists()
         ):
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -105,7 +105,7 @@ class AbstractBuild(TimeStampedEditableModel):
 
     def clean(self):
         # Make sure that ('category__organization', 'os') is unique too
-        if not self.os:
+        if not (self.os and hasattr(self, "category")):
             return
         if (
             load_model('Build')
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -105,12 +105,16 @@ class AbstractBuild(TimeStampedEditableModel):
 
     def clean(self):
         # Make sure that ('category__organization', 'os') is unique too
-        if not (self.os and hasattr(self, "category")):
+        try:
+            category = self.category
+        except ObjectDoesNotExist:
+            return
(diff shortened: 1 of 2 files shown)


openwisp_firmware_upgrader/base/models.py
```diff
@@ -105,7 +105,7 @@ def __str__(self):
 
     def clean(self):
         # Make sure that ('category__organization', 'os') is unique too
-        if not self.os:
+        if not (self.os and hasattr(self, "category")):
```
nemesisdesign, line 108: since `self.category` is going to be accessed anyway below, and `hasattr` causes the attribute to be accessed anyway, I think it may be better to do:

```python
try:
  category = self.category
except ObjectDoesNotExists:
  return
if not self.os:
  return
```
And then below change `self.category` to `category`.

It's a bit longer but more explicit and saves the accessed call which is reused.

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -64,6 +64,18 @@ def test_build_clean(self):
...
+            try:
+                b2 = self._create_build(
+                    os=self.os + '_2', version='0.2', organization=org
+                )
+                b2.category = None
+                b2.full_clean()
+            except ValidationError as e:
+                self.assertIn('category', e.message_dict)
+            else:
+                self.fail('ValidationError not raised when build category is empty')
```
nemesisdesign, line 77: I ensured this test fails without the patch, good! :+1:

nemesisdesign (commented):
Thank you @hanif-ali! One suggestion below.

openwisp_firmware_upgrader/base/models.py
hanif-ali: Yeah, that is much neater. Thanks.

nemesisdesign (changes requested):
@hanif-ali QA checks are failing, see https://openwisp.io/docs/developer/contributing.html#python-code-conventions to find out how to fix them :)

nemesisdesign (approved):
The commit message is not exactly compliant with our guidelines but I can fix it during squash merge, please take a look at it so you can follow the same style next time, BTW the `./run-qa-checks` script will tell you about this problem, so please always run it before pushing.

PR #158, opened 2021-09-14
[enhancement] Use select2 for FirmwareImageInline type field #148

### LINKED ISSUE #148: [ux] Enable Select2 on firemware image model select in admin
Enabling select2 on the firmware image model select in the admin should be just a matter to load the select2 JS (which is already shipped by django) and initialize it on the select.

pandafy:
Closes #148


nemesisdesign (approved):
Great, thanks!

PR #169, opened 2022-01-12
[fix] Fixed restoring deleted device restores related objects #168 

### LINKED ISSUE #168: [bug] Recovering deleted device does not recover related objects
Device firmware and recent firmware upgrade objects are not recovered after restoring a deleted device. These models should be tracked by django-reversion.

**Blockers**
- [ ]  https://github.com/openwisp/openwisp-controller/pull/581

pandafy:
Closes #168

### DIFF
--- a/openwisp_firmware_upgrader/admin.py
+++ b/openwisp_firmware_upgrader/admin.py
@@ -115,11 +115,8 @@ class BuildAdmin(BaseVersionAdmin):
         if model == FirmwareImage:
             kwargs['follow'] = (
                 *kwargs['follow'],
-                'devicefirmware_set',
                 'build',
             )
-        if model == Build:
-            kwargs['follow'] = (*kwargs['follow'],)
         return super().reversion_register(model, **kwargs)
 
     def upgrade_selected(self, request, queryset):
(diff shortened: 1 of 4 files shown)


nemesisdesign (changes requested):
So here's what's going on, very weird.
I have a default credential object created.
I create a new device (after switching to this branch and making sure I have the right openwisp-controller branch activated), I assign it a config, I create also a device firmware.
So the device has config, credentials and firmware.
I delete it the device.
When I go to recover page, I see 2 device credential objects.
Now I do not know if this has to be fixed in openwisp-controller or fw-upgrader, likely controller, But I found this while testing this patch.

Screenshot from 2022-01-13 16-34-40

If I try to restore it I get a generic validation error with no info.
If while restoring I check the "delete" checkbox of the device credentials, the object gets restored.

I tried the same in openwisp-controller right now and it doesn't happen.

Are you sure that this patch isn't overwriting some method that's defined in openwisp-controller and hence some logic which should get executed does not get executed?

okraits:
We also had problems restoring deleted objects at some point, with a meaningless error message showing up. I guess it's because of dependencies to other objects which have been modified or deleted as well.
I haven't looked into this more deeply but I guess it would be beneficial to only create atomic revisions (only one object changed/deleted per revision)

nemesisdesign (approved):
It works, thanks :+1:

PR #186, opened 2022-04-09
add hw tp-link wdr 3600

mikysal78:


openwisp_firmware_upgrader/hardware.py
```diff
@@ -172,6 +172,13 @@
                     'boards': ('TP-LINK TL-WDR4300 v1 (IL)',),
                 },
             ),
+            (
+                'ar71xx-generic-tl-wdr3600-v1-squashfs-sysupgrade.bin',
+                {
+                    'label': 'TP-Link WDR3600 v1 (OpenWRT 19.07 and later)',
```
nemesisdesign, line 178: ar71xx is OpenWrt 19.07 and earlier
ath79 is OpenWrt 19.07 and later

openwisp_firmware_upgrader/hardware.py
```diff
@@ -172,6 +172,13 @@
                     'boards': ('TP-LINK TL-WDR4300 v1 (IL)',),
                 },
             ),
+            (
+                'ar71xx-generic-tl-wdr3600-v1-squashfs-sysupgrade.bin',
```
nemesisdesign, line 176: this should go before WDR 4300, we also need the ath79 equivalent

nemesisdesign (changes requested):
Pointing out a few issues, I am going to patch those.

PR #187, opened 2022-04-11
[change!] Made API url consistent with other OpenWISP modules #185

### LINKED ISSUE #185: [change!] API URL structure is not consistent with other OpenWISP modules
As in OpenWISP Users, where we added the `/users/` prefix to all URLs, here we should make sure the prefix is `/firmware-upgrader/`: the prefix should be the name of the module without `openwisp-`, right now the prefix is only `firmware`.

codesankalp:
Closes #185

### DIFF
--- a/openwisp_firmware_upgrader/api/urls.py
+++ b/openwisp_firmware_upgrader/api/urls.py
@@ -6,7 +6,7 @@ app_name = 'upgrader'
 
 urlpatterns = [
     path(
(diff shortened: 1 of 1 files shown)


nemesisdesign (changes requested):
The resulting URLs do not look right!

Screenshot from 2022-04-11 15-14-07

nemesisdesign (approved):

PR #209, opened 2022-11-18
[change] Try all available DeviceConnection for performing upgrades #207

### LINKED ISSUE #207: [change] Rewrite connect logic to try multiple credentials
Depends on https://github.com/openwisp/openwisp-controller/issues/704.

Update the connection logic to use the new method which will be introduced in https://github.com/openwisp/openwisp-controller/issues/704.

https://github.com/openwisp/openwisp-firmware-upgrader/blob/7192c34d321a095271dab9879b4272e13dc0204c/openwisp_firmware_upgrader/base/models.py#L521-L526

pandafy:
Use "DeviceConnection.get_working_connection" to establish a working connection with the device.

Closes #207

### DIFF
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -530,17 +530,32 @@ class AbstractUpgradeOperation(TimeStampedEditableModel):
 
     def upgrade(self, recoverable=True):
         DeviceConnection = swapper.load_model('connection', 'DeviceConnection')
-        if self.device.deviceconnection_set.count() == 0:
-            self.log_line('No device connection available')
-            return
         try:
             conn = DeviceConnection.get_working_connection(self.device)
-        except NoWorkingDeviceConnectionError:
+        except NoWorkingDeviceConnectionError as error:
+            if error.connection is None:
+                self.log_line('No device connection available')
+                return
+
+            log_template = (
+                'Failed to connect with device using {credentials}.'
+                ' Error: {failure_reason}'
+            )
+            for conn in self.device.deviceconnection_set.select_related('credentials'):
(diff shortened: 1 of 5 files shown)


openwisp_firmware_upgrader/base/models.py
```diff
@@ -518,12 +519,29 @@ def log_line(self, line, save=True):
...
-        if not conn:
+        DeviceConnection = swapper.load_model('connection', 'DeviceConnection')
+        if self.device.deviceconnection_set.count() == 0:
             self.log_line('No device connection available')
             return
+        try:
+            conn = DeviceConnection.get_working_connection(self.device)
+        except NoWorkingDeviceConnectionError:
+            self._recoverable_failure_handler(
+                recoverable, RecoverableFailure('Connection failed')
```
nemesisdesign, line 540: I think we need to explicitly state the credentials which were tried and the failure reason for each of them (one per line).
It's a bit redundant with the info available in the credentials tab but it helps user find this information faster when they're working on upgrades which is better.

Please add a test for this and in the test assert that `DeviceConnection.get_working_connection` is getting called.

nemesisdesign (changes requested):
Good progress Gagan :+1: , see my comment below.

openwisp_firmware_upgrader/base/models.py
pandafy: The `NoWorkingDeviceConnectionError` only contains information about the last tried credential. 


https://github.com/openwisp/openwisp-controller/blob/8481351812f06f9afee8e9936f57a2ce84398a8a/openwisp_controller/connection/base/models.py#L265

nemesisdesign (approved):

PR #216, opened 2022-12-08
[feature] Switched to autocomplete filters #214

### LINKED ISSUE #214: [feature] Switch to autocomplete filters and autocomplete organization fields
We have added support of multi-tenant autocomplete filters in https://github.com/openwisp/openwisp-users/pull/331 

We need to update the following filters here: 
- CategoryAdmin
    - Update organization filter to add autocomplete functionality 
- BuildAdmin
    - Update organization filter to add autocomplete functionality 
    - Update category filter to use MultitenantRelatedOrgFilter
- BatchUpgradeOperationAdmin
    - Update organization filter to add autocomplete functionality 
    - Update category filter to use MultitenantRelatedOrgFilter

Let's also change the organization fields of all the admin classes to use autocomplete fields instead of select fields, which do not scale when the number of organization in the system is high.

Aryamanz29:
Closes #214

### DIFF
--- a/openwisp_firmware_upgrader/admin.py
+++ b/openwisp_firmware_upgrader/admin.py
@@ -20,6 +20,12 @@ from openwisp_controller.config.admin import DeviceAdmin
 from openwisp_users.multitenancy import MultitenantAdminMixin, MultitenantOrgFilter
 from openwisp_utils.admin import ReadOnlyAdmin, TimeReadonlyAdminMixin
 
+from .filters import (
+    BuildCategoryFilter,
+    BuildCategoryOrganizationFilter,
+    CategoryFilter,
+    CategoryOrganizationFilter,
+)
 from .hardware import REVERSE_FIRMWARE_IMAGE_MAP
 from .swapper import load_model
 
@@ -45,8 +51,9 @@ class BaseVersionAdmin(MultitenantAdminMixin, TimeReadonlyAdminMixin, VersionAdm
 @admin.register(load_model('Category'))
 class CategoryAdmin(BaseVersionAdmin):
     list_display = ['name', 'organization', 'created', 'modified']
-    list_filter = [('organization', MultitenantOrgFilter)]
+    list_filter = [MultitenantOrgFilter]
     list_select_related = ['organization']
+    autocomplete_fields = ['organization']
     search_fields = ['name']
     ordering = ['-name', '-created']
 
@@ -80,23 +87,17 @@ class FirmwareImageInline(TimeReadonlyAdminMixin, admin.StackedInline):
         return True
 
 
-class CategoryFilter(MultitenantOrgFilter):
-    multitenant_lookup = 'organization_id__in'
-
-
 @admin.register(load_model('Build'))
 class BuildAdmin(BaseAdmin):
     list_display = ['__str__', 'organization', 'category', 'created', 'modified']
-    list_filter = [
-        ('category__organization', MultitenantOrgFilter),
-        ('category', CategoryFilter),
-    ]
+    list_filter = [CategoryOrganizationFilter, CategoryFilter]
     list_select_related = ['category', 'category__organization']
     search_fields = ['category__name', 'version', 'os']
     ordering = ['-created', '-version']
     inlines = [FirmwareImageInline]
     actions = ['upgrade_selected']
     multitenant_parent = 'category'
+    autocomplete_fields = ['category']
 
     # Allows apps that extend this modules to use this template with less hacks
     change_form_template = 'admin/firmware_upgrader/change_form.html'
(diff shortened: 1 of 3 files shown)


openwisp_firmware_upgrader/admin.py
```diff
@@ -45,8 +51,9 @@ class BaseVersionAdmin(MultitenantAdminMixin, TimeReadonlyAdminMixin, VersionAdm
 @admin.register(load_model('Category'))
 class CategoryAdmin(BaseVersionAdmin):
     list_display = ['name', 'organization', 'created', 'modified']
-    list_filter = [('organization', MultitenantOrgFilter)]
+    list_filter = [MultitenantOrgFilter]
```
nemesisdesign, line 54: This is looking weird to me:
Screenshot from 2022-12-10 13-12-00

nemesisdesign (changes requested):
Almost all good except a detail below.

openwisp_firmware_upgrader/admin.py
Aryamanz29: @nemesisdesign Could you please double check this? because I'm getting the right autocomplete filters, see below:

Screenshot from 2022-12-11 16-11-29

Screenshot from 2022-12-11 16-10-44

nemesisdesign: Thanks for your patience, something was wrong with my setup, all looks good, I am going to ask @pandafy to double check this since he implemented the base code for this, thanks! :+1:

nemesisdesign (dismissed):

openwisp_firmware_upgrader/filters.py
```diff
@@ -0,0 +1,31 @@
...
+Category = load_model('Category')
+
+
+class CategoryFilter(MultitenantRelatedOrgFilter):
+    field_name = 'category'
+    parameter_name = 'category_id'
+    title = _('category')
+
+
+class CategoryOrganizationFilter(MultitenantRelatedOrgFilter):
```
pandafy, line 17:
```suggestion
class CategoryOrganizationFilter(MultitenantOrgFilter):
```

openwisp_firmware_upgrader/filters.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class BuildCategoryFilter(MultitenantRelatedOrgFilter):
+    field_name = 'category'
+    parameter_name = 'build__category'
+    title = _('category')
+    rel_model = Build
+
+
+class BuildCategoryOrganizationFilter(MultitenantRelatedOrgFilter):
```
pandafy, line 29:
```suggestion
class BuildCategoryOrganizationFilter(MultitenantOrgFilter):
```

openwisp_firmware_upgrader/filters.py
```diff
@@ -0,0 +1,31 @@
+from django.utils.translation import gettext_lazy as _
+
+from openwisp_users.multitenancy import MultitenantRelatedOrgFilter
```
pandafy, line 3:
```suggestion
from openwisp_users.multitenancy import MultitenantOrgFilter, MultitenantRelatedOrgFilter
```

pandafy (changes requested):
All organization filters must extend `MultitenantOrgFilter`

openwisp_firmware_upgrader/filters.py
```diff
@@ -0,0 +1,31 @@
...
+
+from openwisp_users.multitenancy import MultitenantOrgFilter
+
+from .swapper import load_model
+
+Build = load_model('Build')
+Category = load_model('Category')
+
+
+class CategoryFilter(MultitenantOrgFilter):
```
nemesisdesign, line 11: This doesn't work as expected, I believe this should inherit MultitenantRelatedOrgFilter.

If you try this filter with a staff user which is org manager but not super admin you should see the list of categories empty even if you create categories for the org the user manages, for example, in my dev env this user has access to 2 orgs and there are existing categories in the DB: 

Screenshot from 2022-12-13 15-53-46

But the filter is not showing any category:
Screenshot from 2022-12-13 15-53-30

nemesisdesign (changes requested):

openwisp_firmware_upgrader/filters.py
```diff
@@ -0,0 +1,31 @@
...
+    parameter_name = 'category_id'
+    title = _('category')
+
+
+class CategoryOrganizationFilter(MultitenantOrgFilter):
+    parameter_name = 'category__organization'
+    rel_model = Category
+
+
+class BuildCategoryFilter(MultitenantOrgFilter):
```
nemesisdesign, line 22: same here as above

nemesisdesign (approved):
Thanks! :+1:

PR #222, opened 2023-01-25
[change] PC Engines apu6 should use x86/64 firmware & apu2 in lower case

kosli:
As I have tested a PC Engines apu6 board, this can be added to the list too.


nemesifier (approved):
Thanks :+1:

PR #226, opened 2023-03-02
[feature] Allowed specifying options for upgrade operations #155

### LINKED ISSUE #155: [feature] Add support sysupgrade options such as -n and --force

pandafy:
Closes #155


openwisp_firmware_upgrader/base/models.py
```diff
@@ -626,6 +630,17 @@ def upgrade(self, recoverable=True):
...
+        super().clean()
+        if self.upgrade_options:
+            # Only store the keys that were set
+            pop_keys = []
+            for key in self.upgrade_options.keys():
+                if not self.upgrade_options[key]:
+                    pop_keys.append(key)
+            for key in pop_keys:
+                self.upgrade_options.pop(key, None)
+
```
pandafy, line 643: @nemesisdesign I want to replace this with proper JSONSchema validation. I am planning to achieve this with the following code 

```python
    def clean(self):
        super().clean()
        self.validate_upgrade_options(self)

    def validate_upgrade_options(self):
        DeviceConnection = swapper.load_model('connection', 'DeviceConnection')
        try:
            conn = DeviceConnection.get_working_connection(self.device)
        except NoWorkingDeviceConnectionError:
            raise ValidationError(_('This device does not have a working connection.'))
        try:
            upgrader_class = app_settings.UPGRADERS_MAP[conn.update_strategy]
            upgrader_class = import_string(upgrader_class)
        except (AttributeError, ImportError):
            raise ValidationError(_('This device does not have a supported upgrader.'))
        try:
            jsonschema.Draft4Validator(upgrader_class.SCHEMA).validate(self.upgrade_options)
        except jsonschema.SchemaError as error:
            raise ValidationError({'upgrade_options': error.message})
```

But, this will make validation very slow and prone to errors. Shall we do schema validation directly at upgrader class?

nemesisdesign: I think the simplest solution to the problem of automatically figuring out which upgrader class to use is the following:

- query the DB for the available device connection, filter only SSH connections
- get the first available SSH connection
  (it doesn't matter if it works or not, because there cannot be connections for two different firmware types but rather two different credentials, eg: one with password and one with SSH key, or different SSH keys, which doesn't make any difference for determining the firmware type)
  this means to not hit the network to test if the connection works or not
- now from the connection you can figure out what upgrader to use `openwisp_firmware_upgrader.settings.UPGRADERS_MAP`

This will make the system extensible because when custom upgrader classes are used are set via the `UPGRADERS_MAP` setting, that is, if I change the default firmware upgrader related to the OpenWrt ssh connection, and the custom upgrader class has a different schema, this schema shall be picked automatically by the system. That's the whole point of this effort.

openwisp_firmware_upgrader/admin.py
```diff
@@ -152,6 +180,12 @@ def upgrade_selected(self, request, queryset):
                 'related_count': len(related_device_fw),
                 'firmwareless_devices': firmwareless_devices,
                 'firmwareless_count': len(firmwareless_devices),
+                'form': form,
+                # TODO: The OpenWrt schema is hard coded here. We need to find
+                # a way to dynamically select schema of the appropriate upgrader.
+                'firmware_upgrader_schema': json.dumps(
+                    OpenWrt.SCHEMA, cls=DjangoJSONEncoder
+                ),
```
nemesisdesign, line 188: So I am counting, this is one.

openwisp_firmware_upgrader/base/models.py
```diff
@@ -626,6 +635,16 @@ def upgrade(self, recoverable=True):
             self.device.devicefirmware.installed = True
             self.device.devicefirmware.save(upgrade=False)
 
+    def clean(self):
+        super().clean()
+        # TODO: The OpenWrt upgrader schema is hard-coded here.
+        # We need to find a way to dynamically select schema
+        # of the correct upgrader class.
+        try:
+            jsonschema.Draft4Validator(OpenWrt.SCHEMA).validate(self.upgrade_options)
```
nemesisdesign, line 644: two.

openwisp_firmware_upgrader/admin.py
```diff
@@ -200,9 +234,30 @@ def has_delete_permission(self, request, obj):
...
+        css = {'all': ['firmware-upgrader/css/upgrade-options.css']}
+
+
+class ReadonlyUpgradeOptionsMixin:
+    @admin.display(description=_('Upgrade options'))
+    def readonly_upgrade_options(self, obj):
+        # TODO: The OpenWrt schema is hard coded here. We need to find
+        # a way to dynamically select schema of the appropriate upgrader.
+        options = []
+        for key, value in OpenWrt.SCHEMA['properties'].items():
```
nemesisdesign, line 247: 3, these lines need to call a function or something which centralizes the logic to figure out which schema to use.

nemesisdesign (changes requested):
Great progress Gagan! I like how it is coming along :+1: 

I took the time to look at the code to give you the best recommendation regarding how to automatically figure out the upgrader class to use to load the schema, read on to find out.

openwisp_firmware_upgrader/admin.py
```diff
@@ -125,33 +143,61 @@ def upgrade_selected(self, request, queryset):
...
         context.update(
             {
                 'title': title,
                 'related_device_fw': related_device_fw,
                 'related_count': len(related_device_fw),
                 'firmwareless_devices': firmwareless_devices,
                 'firmwareless_count': len(firmwareless_devices),
+                'form': form,
+                # TODO: The OpenWrt schema is hard coded here. We need to find
+                # a way to dynamically select schema of the appropriate upgrader.
```
nemesisdesign, line 197: Is this comment up to date? Doesn't look like it.

openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
```diff
@@ -72,7 +72,7 @@ def mocked_exec_upgrade_success(command, exit_codes=None, timeout=None):
 
 
 def mocked_sysupgrade_failure(command, exit_codes=None, timeout=None):
-    if command.startswith(f'{OpenWrt._SYSUPGRADE} -v -c '):
+    if command.startswith(f'{OpenWrt._SYSUPGRADE} -v '):
```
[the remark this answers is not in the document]
nemesisdesign: I am insisting: shouldn't this line remain unchanged? Why is it changing?

openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
```diff
@@ -56,7 +56,7 @@ def mocked_exec_upgrade_success(command, exit_codes=None, timeout=None):
     }
     if command.startswith(f'{_sysupgrade} --test /tmp/openwrt-'):
         return defaults
-    if command.startswith(f'{_sysupgrade} -v -c /tmp/openwrt-'):
```
nemesisdesign, line 59: Same here

nemesisdesign (changes requested):
Great progress @pandafy! :+1: :+1: 

Here's a few problems I found during testing. It is not exhaustive, we'll need to perform more testing.

### Upgrade options shown when device firmwar image is set to None

In Device Firmware inline, if you change the firmware image from a selected image to none ("--------"), the upgrade options appear, but this doesn't make sense, does it?

### Device Firmware CSS needs a bit more love

Screenshot from 2023-03-19 19-53-32

I recommend adding a bit more spacing between the first option and the title "UPGRADE OPTIONS" (maintain padding consistent please), while after the last option, I recommend ensuring there's a border like all other rows have (we should treat the entire option section as a row, so consistent padding/margin and border before the next row).

### Testing

Looking forward to see the small test coverage issue fixed :blush:

nemesisdesign (changes requested):
Looking very good, can you raise a validation error if both -n and -c or -n and -o are enabled, at the same time? The option -n is mutually esclusive with -c or -o.

nemesisdesign (commented):
Looks good! I will keep testing it a bit more in the coming days.

PR #228, opened 2023-03-17
Add Support for CJ-Hello HYC-G920 Device

momothefox:
Add Support for CJ-Hello HYC-G920 Device

### DIFF
--- a/openwisp_firmware_upgrader/hardware.py
+++ b/openwisp_firmware_upgrader/hardware.py
@@ -56,6 +56,13 @@ OPENWRT_FIRMWARE_IMAGE_MAP.update(
                 {'label': 'GL.iNet GL-MT300N-V2', 'boards': ('GL-MT300N-V2',)},
             ),
             (
+                'hanyang_hyc-g920-squashfs-sysupgrade.bin',
+                {
+                    'label': 'CJ-Hello HYC-G920',
+                    'boards': ('CJ-Hello HYC-G920',),
+                },
+            ),
+            (
                 'mvebu-cortexa9-linksys_wrt1900acs-squashfs-sysupgrade.img',
                 {'label': 'Linksys WRT1900ACS', 'boards': ('Linksys WRT1900ACS',)},
             ),


openwisp_firmware_upgrader/hardware.py
```diff
@@ -55,6 +55,13 @@
                 'ramips-mt76x8-gl-mt300n-v2-squashfs-sysupgrade.bin',
                 {'label': 'GL.iNet GL-MT300N-V2', 'boards': ('GL-MT300N-V2',)},
             ),
+            (
+                'hanyang_hyc-g920-squashfs-sysupgrade.bin',
```
nemesisdesign, line 59: can you provide a link to the downloads.openwrt.org build for this device please? The architecture seems missing. The alphabetical order is not right in the list but I can fix that.

nemesisdesign (changes requested):

openwisp_firmware_upgrader/hardware.py
momothefox: @nemesisdesign  this device is not yet officially supported. i have have my own  build for it, 
i tried to add support for it on openwrt repo but with no success. 
they prefer to go with original u boot while the device is hacked with breedweb. 
you can ignore this until i can get a device with original boot to fulfill openwrt requirements

nemesisdesign: I see, for these cases we have this setting which you can use: https://github.com/openwisp/openwisp-firmware-upgrader#openwisp_custom_openwrt_images
We cannot include in the standard list devices which are not supported officially because we would fill the list with items which are not used by anyone except 1 user.

momothefox: yes i understand, by the way the device is 4 core 256 ram 16mb spi, it is just a matter of time i am waiting for it in shipping 
[https://www.hanyangdgt.com/FILE/CustomerService/Cs4/HYC-G920_KTC/HYC-G920%20%EC%A0%9C%ED%92%88%EC%86%8C%EA%B0%9C%EC%84%9C.pdf](url)

i rushed this up i already did added it to my list as much as dw02-412h that i added earlier to the project

momothefox: @nemesisdesign is you could consider this, it would be great, i am using it now  along with openwisp, by joining all networks from openwisp server.. so openwisp controller can reach all different clients in separated networks, 
and it is zero configuration nearly. 
[https://docs.zerotier.com/openapi/servicev1.json](url)

i am not a dev. but i think it would match with openwisp fantastic operations.

nemesisdesign: @momothefox as soon as there's going to be enough people interested gathering resources to implement this we'll work on it: https://github.com/openwisp/openwisp-controller/issues/604 .

Issue #206, opened 2022-10-26
[feature] REST API is missing endpoints for DeviceFirmware


nemesisdesign:
Right now it is not possible to do the following:

- CRUD operations of DeviceFirmware objects, eg: `/api/v1/firmware-upgrader/device/[id]/firmware/`
  - When a device firmwre is created or changed (which creates an upgrade operation) we should return the upgrade operation ID to allow clients to check the state of the upgrade
- Read UpgradeOperation objects, eg: `/api/v1/firmware-upgrader/device/[id]/upgrade-operation/`

tspare:
I would like to take a shot on this one but would like the expert's opinion to see whats the right approach.  Any comments are greatly appreciated.
On the (/api/v1/firmware-upgrader/device/[id]/firmware/) endpoint, I think we could go 2 ways:
1.  Follow the pattern in DeviceUpgradeOperationInline and do a search of the  UpgradeOperations and return the UpgradeOperation ID of last XXX upgrades done for the device
2.  Build a separate model per device that saves all the UpgradeOperation done and return the IDs from that model

I am not sure I am on the right track, any suggestion as to the best approach?

PR #229, opened 2023-04-04
 [feature] REST API is missing endpoints for DeviceFirmware #206

tspare:
[feature] REST API is missing endpoints for DeviceFirmware #206

The code adds three endpoints:

1.  upgrade-operation/ - List all the upgrade operation in the system
2.  upgrade-operation/<id>/ - Display detail of a particular upgrade id
3.  device/<id>/upgrade-operation/ - return a list of upgrade-operation ids for a gven device id

### DIFF
--- a/openwisp_firmware_upgrader/api/serializers.py
+++ b/openwisp_firmware_upgrader/api/serializers.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
 
 from openwisp_users.api.mixins import FilterSerializerByOrgManaged
@@ -10,6 +11,7 @@ Build = load_model('Build')
 Category = load_model('Category')
 FirmwareImage = load_model('FirmwareImage')
 UpgradeOperation = load_model('UpgradeOperation')
+DeviceFirmware = load_model("DeviceFirmware")
 
 
 class BaseMeta:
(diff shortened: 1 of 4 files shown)


nemesisdesign (changes requested):
Thanks for contributing!

The main requirement of the feature is not satisfied:

- CRUD operations of DeviceFirmware objects, eg: /api/v1/firmware upgrader/device/[id]/firmware/ 
  When a device firmwre is created or changed (which creates an upgrade operation) we should return the upgrade operation ID to allow clients to check the state of the upgrade

We will need to add tests and docs (add these endpoints to the README).

tspare:
> Thanks for contributing!
> 
> The main requirement of the feature is not satisfied:
> 
>     * CRUD operations of DeviceFirmware objects, eg: /api/v1/firmware upgrader/device/[id]/firmware/
>       When a device firmwre is created or changed (which creates an upgrade operation) we should return the upgrade operation ID to allow clients to check the state of the upgrade
> 
> 
> We will need to add tests and docs (add these endpoints to the README).

I updated the code to add the changes.  However, I am not sure if you want the upgrade operation to be saved in to the DeviceFirmware model or just return it as a separate field as part of the PUT request.   I left it as a separate field that is returned as part of the PUT request.  Please let me know if this is what you had in mind or if I am out on the left field.

nemesisdesign:
Thanks for your effort @tspare, we improved upon your work and merged this in https://github.com/openwisp/openwisp-firmware-upgrader/pull/231.

tspare:
You are welcome.  Glad I could help.

PR #231, opened 2023-05-04
[feature] Added missing endpoints for DeviceFirmware and UpgradeOperation #206

### LINKED ISSUE #206: [feature] REST API is missing endpoints for DeviceFirmware

Aryamanz29:
**Blockers**

- [x] https://github.com/openwisp/openwisp-firmware-upgrader/pull/234
- [x] https://github.com/openwisp/openwisp-users/pull/352

- [x] This PR adds four endpoints:
1.  `upgrade-operation/`
    - List all the upgrade operation in the system
   
2.  `upgrade-operation/<id>/`
    - Display detail of a particular upgrade id

3.  `device/<device_pk>/upgrade-operation/`
    - return a list of upgrade-operation ids for a given device id

4.  `device/<device_pk>/firmware/`

    - **PUT:** if DeviceFirmware doesn't exist, create a new object else
            update the image of the DeviceFirmware. 
			Returns the DeviceFirmware object and if an upgrade operation
            was created, also return the id of the UpgradeOperation.

    - **GET:** return the DeviceFirmware object,
 
- [x]  Added endpoints tests.
- [x] Updated documentation.

Closes #206

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -473,19 +473,16 @@ List mass upgrade operations
 
 **Available filters**
 
-You can filter a list of batch upgrade operations
-based on their build using the ``build_id``.
+The list of batch upgrade operations provides the following filters:
+
+- ``build`` (Firmware build ID)
+- ``status`` (One of: idle, in-progress, success, failed)
+
+Here's a few examples:
 
 .. code-block:: text
 
     GET /api/v1/firmware-upgrader/batch-upgrade-operation/?build={build_id}
-
-You can filter a list of batch upgrade
-operations based on their status using the
-``status`` (e.g. idle, in-progress, success, failed).
-
-.. code-block:: text
-
     GET /api/v1/firmware-upgrader/batch-upgrade-operation/?status={status}
 
 Get mass upgrade operation detail
@@ -504,25 +501,18 @@ List firmware builds
 
 **Available filters**
 
-You can filter a list of firmware builds based
-on their category using the ``category_id``.
+The list of firmware builds provides the following filters:
+
+- ``category`` (Firmware category ID)
+- ``version``  (Firmware build version)
+- ``os`` (Firmware build os identifier)
+
+Here's a few examples:
 
 .. code-block:: text
 
     GET /api/v1/firmware-upgrader/build/?category={category_id}
-
-You can filter a list of firmware builds based
-on their version using the ``version`` (e.g. 1.2.x).
-
-.. code-block:: text
-
     GET /api/v1/firmware-upgrader/build/?version={version}
-
-You can filter a list of firmware builds based
-on their os identifier using the ``os``.
-
-.. code-block:: text
-
     GET /api/v1/firmware-upgrader/build/?os={os}
 
 Create firmware build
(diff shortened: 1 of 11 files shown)


nemesisdesign (commented):
Looks good, I will spend more time testing this tomorrow!

README.rst
```diff
@@ -619,6 +619,65 @@ Delete a firmware category
...
+.. code-block:: text
+
+    GET /api/v1/firmware-upgrader/upgrade-operation/{id}
+
+List device upgrade operations
+##############################
+
+.. code-block:: text
+
+    GET /api/v1/firmware-upgrader/device/{device_pk}/upgrade-operation/
```
nemesisdesign, line 641: can we use `device_id` for consistency with the rest of the document please?

nemesisdesign (changes requested):
Here's some feedback from the first round of testing.
I did a quick test of the different endpoints which seem to work quite well.
Below I have some minor change requests.

**Upgrade Operation List:**
- can you allow to filter by org and org_slug?
- can you allow to filter by device ID (can we avoid generating the select with the entire device list in the UI to avoid this to fail on large instances?)
- allow filter by image ID (can we avoid generating the select with the entire image list in the UI to avoid this to fail on large instances?)
- allow filter by status

**Device Upgrade Operation List:**
- allow filter by status

I will do another round of acid testing and code review tomorrow.

README.rst
```diff
@@ -471,6 +471,23 @@ List mass upgrade operations
 
     GET /api/v1/firmware-upgrader/batch-upgrade-operation/
 
+**Available filters**
+
+You can filter a list of batch upgrade operations
+based on their build using the ``build_id``.
```
nemesisdesign, line 477: It's not `build_id`, it's `build`.
I wouldn't use e.g. to list the possible values of the filter, because e.g. means "example", while here we're listing the possible values.

I would rewrite it as follows:

Bacth upgrade operations can be filtered by build ID and status (one of: idle, in-progress, success, failed):

.. code-block:: text

    GET /api/v1/firmware-upgrader/batch-upgrade-operation/?build={build_id}
    GET /api/v1/firmware-upgrader/batch-upgrade-operation/?status={status}

Please let's make this docs text less verbose, apply the same suggestion to all other sections.

nemesisdesign (changes requested):
It's looking good, I have some remaining doubts below.

README.rst
Aryamanz29: Sure.

README.rst
```diff
@@ -619,6 +669,113 @@ Delete a firmware category
...
+
+    GET /api/v1/firmware-upgrader/upgrade-operation/?image={image_id}
+
+You can filter a list of upgrade
+operations based on their status using the
+``status`` (e.g. in-progress, success, failed, aborted).
+
+.. code-block:: text
+
+    GET /api/v1/firmware-upgrader/upgrade-operation/?status={status}
```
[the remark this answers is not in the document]
Aryamanz29: Yes that looks good, thanks :)

openwisp_firmware_upgrader/base/models.py
```diff
@@ -321,13 +321,17 @@ def image_has_changed(self):
...
             self.installed = False
             super().save(*args, **kwargs)
-            self.create_upgrade_operation(batch, upgrade_options=upgrade_options or {})
+            upgrade_operation = self.create_upgrade_operation(
+                batch, upgrade_options=upgrade_options or {}
+            )
         else:
             super().save(*args, **kwargs)
         self._update_old_image()
+        return upgrade_operation
```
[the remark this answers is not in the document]
Aryamanz29: You are correct, we don't actually need this change. Thank you for bringing it to my attention!

nemesisdesign (approved):
Thanks! :+1:

Issue #233, opened 2023-05-18
[bug] Upgrading using -n causes host key error


nemesifier:
When upgrading an OpenWrt device using the `-n` option (do not save configuration over reflash), this may happen when the system tries to reconnect:

```
Trying to reconnect to device at 172.30.131.203 (attempt n.8)...
Device not reachable yet, (Host key for server '172.30.XXX.XXX' does not match: got 'AAAAB3N.....').
```

We should devise a generic mechanism to indicate to the upgrader that it shall auto add the new host key whenever the config of the device is lost on purpose.

nemesifier:
Fixed by https://github.com/openwisp/openwisp-firmware-upgrader/pull/236

PR #236, opened 2023-05-30
[fix] Fixed issues in reconnecting to device after firmware upgrade #235

### LINKED ISSUE #235: [bug] No valid IP addresses to initiate connections found
Recent versions of OpenWISP Controller can raise a ValueError exception which is not caught by this module.
When this happens, instead of continuing to retry, the retry mechanism stops, eg:

```
Trying to reconnect to device at 172.16.0.32 (attempt n.5)...
Device not reachable yet, (timed out).
retrying in 20 seconds...
Trying to reconnect to device at (attempt n.6)...
No valid IP addresses to initiate connections found
```

This looks like a bug. The system must keep retrying until the max amount of retries are reached, in that case the last line of the log would look different.

nemesisdesign:
The test is misssing.

Fixes #235

This may fix also #233 (needs manual testing to confirm).

### DIFF
--- a/openwisp_firmware_upgrader/upgraders/openwrt.py
+++ b/openwisp_firmware_upgrader/upgraders/openwrt.py
@@ -9,6 +9,7 @@ from django.utils.translation import gettext_lazy as _
 from paramiko.ssh_exception import NoValidConnectionsError, SSHException
 
 from openwisp_controller.connection.connectors.openwrt.ssh import OpenWrt as BaseOpenWrt
+from openwisp_controller.connection.exceptions import NoWorkingDeviceConnectionError
 
 from ..exceptions import (
     FirmwareUpgradeOptionsException,
@@ -408,27 +409,28 @@ class OpenWrt(BaseOpenWrt):
         handle cases in which the IP has changed
         """
         self.connection.device.refresh_from_db()
-        self.connection.refresh_from_db()
-        del self.connection.connector_instance
+        self.connection = self.connection.get_working_connection(self.connection.device)
         self.addresses = self.connection.get_addresses()
         self._params = self.connection.connector_instance._params
         self.shell = self.connection.connector_instance.shell
 
     def _write_checksum(self, checksum):
         for attempt in range(1, self.RECONNECT_MAX_RETRIES + 1):
-            self._refresh_addresses()
-            addresses = ', '.join(self.addresses)
-            self.log(
-                _(
-                    'Trying to reconnect to device at {addresses} (attempt n.{attempt})...'.format(
-                        addresses=addresses, attempt=attempt
-                    )
-                ),
-                save=False,
-            )
             try:
+                self._refresh_addresses()
+                addresses = ', '.join(self.addresses)
+                self.log(
+                    _(
+                        'Trying to reconnect to device at {addresses} (attempt n.{attempt})...'.format(
+                            addresses=addresses, attempt=attempt
+                        )
+                    ),
+                    save=False,
+                )
                 self.connection.connect()
-            except (NoValidConnectionsError, socket.timeout, SSHException) as error:
+            except (NoWorkingDeviceConnectionError, NoValidConnectionsError, socket.timeout, SSHException) as error:
+                if not str(error):
+                    error = _('connection failed')
                 self.log(
                     _(
                         'Device not reachable yet, ({0}).\n'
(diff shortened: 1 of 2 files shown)


openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -408,24 +417,31 @@ def _refresh_addresses(self):
...
+                addresses = ', '.join(self.addresses)
+                self.log(
+                    _(
+                        'Trying to reconnect to device at {addresses} (attempt n.{attempt})...'.format(
+                            addresses=addresses, attempt=attempt
+                        )
+                    ),
+                    save=False,
+                )
                 self.connect()
```
pandafy, line 436: While performing manual testing locally, I got this log for the device 

```
Connection successful, starting upgrade...
Image checksum file not found, proceeding with the upload of the new image...
Sysupgrade test passed successfully, proceeding with the upgrade operation...
Upgrade operation in progress...
Commencing upgrade. Closing all shell sessions.
Image metadata not found
Reading partition table from bootdisk...
Reading partition table from image...
Partition layout has changed. Full image will be written.

SSH connection closed, will wait 150 seconds before attempting to reconnect...
Device not reachable yet, (connection failed).
retrying in 30 seconds...
Device not reachable yet, (connection failed).
retrying in 30 seconds...
Trying to reconnect to device at 192.168.56.2 (attempt n.3)...
Connected! Writing checksum file to /etc/openwisp/firmware_checksum
Upgrade completed successfully.
``` 

See `Trying to reconnect to device at 192.168.56.2 (attempt n.3)...` shows third attempt. This is because in previous two attempts the device was unreachable and `self._refresh_addresses()` raised `NoWorkingDeviceConnectionError` which led to the logs getting skipped. 

How can we get updated addresses without initiating a connection?

pandafy:
I also tested #233 locally and this code mitigates that problem.

openwisp_firmware_upgrader/upgraders/openwrt.py
nemesisdesign, line 436: shouldn't we update this too?

nemesisdesign: what about:

```python
```python
def _log(addresses, attempt):
    self.log(
        _(
            f'Trying to reconnect to device at {addresses} '
            '(attempt n.{attempt})...'.format(
                addresses=addresses,
                attempt=attempt
            )
        ),
        save=False,
    )
try:
    self._refresh_addresses()
except NoWorkingDeviceConnectionError as error:
    _log(error.connection.addresses, attempt)
try:
    _log(', '.join(self.addresses), attempt)
    self.connect()
except (NoValidConnectionsError, socket.timeout, SSHException) as error:
    # etc...
```

pandafy: I added the following wrapper methods so we don't need to reference the connection object again and again 

```
https://github.com/openwisp/openwisp-firmware-upgrader/blob/82ade68a1ba8c6cc988dc2a73e577e9abcf6d0e1/openwisp_firmware_upgrader/upgraders/openwrt.py#L123-L130
```

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -408,24 +415,30 @@ def _refresh_addresses(self):
...
-            self.log(
-                _(
-                    'Trying to reconnect to device at {addresses} (attempt n.{attempt})...'.format(
-                        addresses=addresses, attempt=attempt
-                    )
-                ),
-                save=False,
-            )
             try:
-                self.connect()
```
pandafy, line 427: We do not need to perform the connect operation here. `self.connection.get_working_connection` will open the SSH connection.

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -408,24 +415,30 @@ def _refresh_addresses(self):
...
-            except (NoValidConnectionsError, socket.timeout, SSHException) as error:
```
pandafy, line 428: We no longer need to catch these exceptions here. These are handled by DeviceConnection model 

https://github.com/openwisp/openwisp-controller/blob/6adafe3d84763c9e942ccceb6f4c006157e50208/openwisp_controller/connection/base/models.py#L345-L357

Issue #208, opened 2022-11-15
[Feature] Enable cancelation of stuck "in progress" mass firmware updates


tricksol:
On the mass update history tab where you see the running updates, there should be a button to cancel updates that seem to be stuck.

kosli:
I have run into the same "issue" :-)

nemesisdesign:
This is a rare corner case which shouldn't happen thanks to `OPENWISP_FIRMWARE_UPGRADER_TASK_TIMEOUT` ([see relevant section in the readme](https://github.com/openwisp/openwisp-firmware-upgrader#openwisp_firmware_upgrader_task_timeout)).

It can happen though, if while an upgrade is in progress, a cold shut down is performed.

Now, how can we deal with this is the question.

Allow canceling or abort an in-progress upgrade can create other problems and more corner cases to handle.
The upgrade can be canceled if the firmware is not being flashed yet, but once the flashing starts, there's no point in aborting, it's too late.

I think the problem here is that while an upgrade is in progress users cannot do new upgrades, which is very annoying.
A quick fix is to open the django shell and execute this SQL query, just make sure there's no other real upgrade in progress!

```python
UpgradeOperation.objects.filter(status='in-progress').update(status='failed')
```

A definitive solution could be the following: when a new upgraded operation is created if we detect another one in progress, let's check if it's old enough to be timed out, if it is, flag it as failed automatically and move on.

mips171:
Hey, I opened a draft PR, haven't tested it yet though as I don't have my env set up. https://github.com/openwisp/openwisp-firmware-upgrader/pull/238/files

nemesifier:
I think we'll need a mechanism to do this on single upgrades and mass upgrades, we'll need to discuss this because it's delicate: cancelling an upgrade after the sysupgrade command has been sent is probably not a good idea, as the upgrade has started and cannot be cancelled, so we need the system to be aware of this. At the same time, if an upgrade operation is stuck after the sysupgrade command has been called, we need a way to terminate it: we could verify whether the upgrade was performed and then based on the result of this verification we can either flag it as successful or failed.

PR #238, opened 2023-06-09
[fix] Handle upgrade operation is in progress

mips171:
Implemented according to https://github.com/openwisp/openwisp-firmware-upgrader/issues/208#issuecomment-1585126524 `if we detect another one in progress, let's check if it's old enough to be timed out, if it is we assume this corner case is happening, flag it as failed automatically and move on.`

Fixes https://github.com/openwisp/openwisp-firmware-upgrader/issues/208

### DIFF
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import os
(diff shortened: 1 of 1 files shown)


nemesisdesign (commented):
We'll need to add some tests to this PR to get it merged.

PR #247, opened 2023-09-08
[fix] Added handling for sysupgrade returning non-zero exit code #246

### LINKED ISSUE #246: [bug] Sysupgrade returning non zero exit code some times but upgrade is performed anyway
In some cases, for some unknown reason, the sysupgrade command returns a non zero exit code and the upgrade is flagged as failed, but it is carried out anyway, I believe we need to add a workaround in openwisp to recognize this false positive and ignore it.

Sample output:

```
Command failed: ubus call system sysupgrade { "prefix": "\/tmp\/root", "path": "\/tmp\/<file-name-here>", "backup": "\/tmp\/sysupgrade.tgz", "command": "\/lib\/upgrade\/do_stage2", "options": { "save_partitions": 1 } } (Connection failed)
```

pandafy:
Fixes #246

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
+++ b/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
@@ -134,6 +134,22 @@ def mocked_exec_upgrade_memory_aborted(
     return mocked_exec_upgrade_memory_success(command, exit_codes=None, timeout=None)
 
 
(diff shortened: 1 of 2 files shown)


nemesifier (approved):
Looks good, let's test a bit more before merging!

nemesifier (approved):

PR #255, opened 2024-05-14
[fix] Allow longer firmware filenames #241

### LINKED ISSUE #241: Firmware filename is limited to 100 characters
Hi, I'm compiling my own openWRT firmware for a Xiaomi Mi Router 4a Gigabyte edition, but the generated firmware can't be upload to the openWISP server because the limit of firmware filename: `ramips-mt7621-xiaomi_mi-router-4a-gigabit-squashfs-sysupgrade.bin`

image

There are some config to change this or this is a simple limit of model on django migrations?

I'm using the imagebuilder ansible playbook and tested directly on the site too.

praptisharma28:
Fixes #241


.stylelintrc.json
pandafy, line 1: Please send these changes in a separate PR.

openwisp_firmware_upgrader/migrations/0010_alter_firmwareimage_file.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="firmwareimage",
+            name="file",
+            field=private_storage.fields.PrivateFileField(
+                max_length=255,
+                storage=private_storage.storage.files.PrivateFileSystemStorage(
+                    base_url="/firmware/"
```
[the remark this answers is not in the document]
praptisharma28: Done @pandafy

.stylelintrc.json
praptisharma28: @pandafy I removed, should I open an issue for this then?

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -483,3 +484,21 @@ def test_delete_firmware_image_file(self):
...
             image.delete()
+
+    def test_long_filename_upload(self):
+        build_object = self._create_build()
+
+        # Created a filename longer than 100 characters
+        long_filename = "abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcd"
+        "efghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstabcd"
+        "efghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwx"
+        "zabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghiuvwxyz.bin"
```
pandafy, line 495: You can make this concise by dynamically generating the string like so 

```
long_filename = 'a' * 256
```

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -483,3 +484,21 @@ def test_delete_firmware_image_file(self):
             file_storage_backend.delete(file_name)
             self.assertNotEqual(image.file, None)
             image.delete()
+
+    def test_long_filename_upload(self):
```
pandafy, line 488: We need two sub-tests - one would check the filename is within the limits and the other one would check the other way.

pandafy (changes requested):
Please check my comments, then we would be good to go.

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -483,3 +484,33 @@ def test_delete_firmware_image_file(self):
...
+
+        dummy_file = SimpleUploadedFile(long_filename, b"file_content")
+        firmware_image = FirmwareImage(
+            build=build_object,
+            file=dummy_file,
+            type='factory',
+        )
+        self.assertIsNotNone(firmware_image.pk)
+        expected_filename = long_filename
+        self.assertEqual(firmware_image.file.name, expected_filename)
```
pandafy, line 501: I would still call `fimrware_image.full_clean()` here to ensure that validation is not failing.

Just for the sake of readability, I would wrap that in try-except like so 

```suggestion
        try:
            firmware_image.full_clean()
        except ValidationError: 
            self.fail('Unexpected ValidationError raised on running full_clean')
```

Feel free to improve the fail message.

pandafy (changes requested):

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -483,3 +484,38 @@ def test_delete_firmware_image_file(self):
...
+        long_filename = "a" * 256
+
+        dummy_file = SimpleUploadedFile(long_filename, b"file_content")
+        firmware_image = FirmwareImage(
+            build=build_object,
+            file=dummy_file,
+            type='factory',
+        )
+        with self.assertRaises(ValidationError):
+            firmware_image.full_clean()
```
pandafy, line 521: We should ensure that the ValidationError is raised because of the filename and not any other field.

nemesifier: @praptisharma28, look at other test for references, eg:
https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/tests/test_models.py#L55-L61

openwisp_firmware_upgrader/base/models.py
```diff
@@ -211,6 +212,8 @@ class AbstractFirmwareImage(TimeStampedEditableModel):
         upload_to=get_build_directory,
         max_file_size=app_settings.MAX_FILE_SIZE,
         storage=app_settings.PRIVATE_STORAGE_INSTANCE,
+        max_length=255,
+        validators=[MaxLengthValidator(255)],
```
nemesifier, line 216: Isn't this redundant with the line above?

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -31,6 +32,21 @@ class TestModels(TestUpgraderMixin, TestCase):
...
+        build1 = self._create_build(category=category, version='0.1')
+        validlength = self.MAX_FILENAME_LENGTH
+        file = SimpleUploadedFile('a' * validlength, b'')
+        image = self._create_firmware_image(
+            build=build1, type=self.TPLINK_4300_IMAGE, file=file
+        )
+        try:
+            image.full_clean()
+        except ValidationError:
+            self.fail("Filename length exceeds the limit of 255 characters")
```
nemesifier, line 49: As discussed, creating a file with a name longer than 255 chars on Linux is not possible.
Moreover, we trust that the django model logic which caps the max length of a field works, so we don't need to test it again.

nemesifier (changes requested):

nemesifier (approved):

PR #299, opened 2025-04-11
[feature] Added device recognition mechanism

### LINKED ISSUE #165: [feature] Add device recognition mechanism
Due to possible misconfiguration or bugs, in theory it's possible that the firmware upgrader could reflash the wrong device.

Since this is theoretically possible, we should add a mechanism which would be able to spot this case and prevent it from happening.

I immagine it as a method, which would have to be adapted for each upgrader backend, on OpenWRT we could simply check the uuid of the device in /etc/config/openwisp matches the UUID of the device currently being upgraded.

If the method spots this, it should raise an `UpgradeAborted` error but we must ensure the reason is clearly communicated to the user.

youhaveme9:
## Checklist

Closes #165

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
+++ b/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
@@ -245,9 +245,10 @@ class TestOpenwrtUpgrader(TestUpgraderMixin, TransactionTestCase):
         self.assertTrue(device_conn.is_working)
         self.assertEqual(upgrade_op.status, 'aborted')
         self.assertEqual(exec_command.call_count, 1)
+        uuid = '93e76d30-8bfd-4db1-9a24-9875098c9e61'
         lines = [
             'Connection successful, starting upgrade...',
-            f'Device UUID mismatch: expected {device_fw.device.pk}, found 93e76d30-8bfd-4db1-9a24-9875098c9e61 in device configuration',
+            f'Device UUID mismatch: expected {device_fw.device.pk}, found {uuid} in device configuration',
         ]
         for line in lines:
             self.assertIn(line, upgrade_op.log)
--- a/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
+++ b/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
@@ -90,7 +90,10 @@ def mocked_exec_uuid_mismatch(command, exit_codes=None, timeout=None):
 
 def mocked_exec_uuid_not_found(command, exit_codes=None, timeout=None):
     if command == 'uci get openwisp.http.uuid':
-        return ['', 1]  # Return empty output with exit code 1 to simulate UUID not found
+        return [
+            '',
+            1,
+        ]  # Return empty output with exit code 1 to simulate UUID not found
     return mocked_exec_upgrade_success(command, exit_codes, timeout)
 
 
--- a/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
+++ b/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
@@ -72,6 +72,12 @@ def mocked_exec_upgrade_success(command, exit_codes=None, timeout=None):
         raise CommandFailedException()
 
 
+def mocked_exec_uuid_mismatch(command, exit_codes=None, timeout=None):
+    if command == 'uci get openwisp.http.uuid':
+        return ['different-uuid', 0]
+    return mocked_exec_upgrade_success(command, exit_codes, timeout)
+
+
 def mocked_sysupgrade_failure(command, exit_codes=None, timeout=None):
     if command.startswith(f'{OpenWrt._SYSUPGRADE} -v -c'):
         raise CommandFailedException(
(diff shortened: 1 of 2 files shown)


youhaveme9:
```
ERROR: Could not find a version that satisfies the requirement Pillow~=11.1.0 (from django-loci) (from versions: 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7.0, 1.7.1, 1.7.2, 1.7.3, 1.7.4, 1.7.5, 1.7.6, 1.7.7, 1.7.8, 2.0.0, 2.1.0, 2.2.0, 2.2.1, 2.2.2, 2.3.0, 2.3.1, 2.3.2, 2.4.0, 2.5.0, 2.5.1, 2.5.2, 2.5.3, 2.6.0, 2.6.1, 2.6.2, 2.7.0, 2.8.0, 2.8.1, 2.8.2, 2.9.0, 3.0.0, 3.1.0rc1, 3.1.0, 3.1.1, 3.1.2, 3.2.0, 3.3.0, 3.3.1, 3.3.2, 3.3.3, 3.4.0, 3.4.1, 3.4.2, 4.0.0, 4.1.0, 4.1.1, 4.2.0, 4.2.1, 4.3.0, 5.0.0, 5.1.0, 5.2.0, 5.3.0, 5.4.0, 5.4.1, 6.0.0, 6.1.0, 6.2.0, 6.2.1, 6.2.2, 7.0.0, 7.1.0, 7.1.1, 7.1.2, 7.2.0, 8.0.0, 8.0.1, 8.1.0, 8.1.1, 8.1.2, 8.2.0, 8.3.0, 8.3.1, 8.3.2, 8.4.0, 9.0.0, 9.0.1, 9.1.0, 9.1.1, 9.2.0, 9.3.0, 9.4.0, 9.5.0, 10.0.0, 10.0.1, 10.1.0, 10.2.0, 10.3.0, 10.4.0)
ERROR: No matching distribution found for Pillow~=11.1.0
```
build is failing while installing dependencies
is it due to recent changes in django-loci?

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,40 @@ def upload(self, image_file, remote_path):
...
             raise RecoverableFailure(str(e))
 
+    def _verify_device_uuid(self):
+        """
+        Verifies that the UUID of the device being upgraded matches
+        the UUID in the device's configuration
+        """
+        device_uuid = self.upgrade_operation.device.pk
+        # Get UUID from device's openwisp config
+        output, exit_code = self.exec_command(
```
nemesifier, line 165: Please add a test in which this method is mocked and returns a different UUID and verify that `UpgradeAborted` is raised as expected.

nemesifier (changes requested):

openwisp_firmware_upgrader/upgraders/openwrt.py
youhaveme9: I'll add

pandafy (changes requested):
This seems to be on the right track. The existing tests are failing because they do not expect `exec_command` to be called with `uci get openwisp.http.uuid`. 

We will need to update the mocks to include this command.

https://github.com/openwisp/openwisp-firmware-upgrader/blob/9e2f4dba3dc3d1a263f1628e82c5b42901f79e90/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py#L40-L72

openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
```diff
@@ -34,6 +34,14 @@ def mocked_exec_upgrade_not_needed(command, exit_codes=None):
         f'test -f {OpenWrt.CHECKSUM_FILE}': ['', 0],
         f'cat {OpenWrt.CHECKSUM_FILE}': [TEST_CHECKSUM, 0],
     }
+    # Handle the UUID command dynamically
+    if command == 'uci get openwisp.http.uuid':
+        # Return the device UUID that matches the one in the test
+        from openwisp_firmware_upgrader.models import DeviceFirmware
```
pandafy, line 40: In OpenWISP, we don't import the model directly. We use the swapper module to load the model. 

The `DeviceFirmware` model is already [loaded globally](https://github.com/openwisp/openwisp-firmware-upgrader/blob/78652eb953c8e59363764b65f9a6ba548b63673e/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py#L24). You don't need to do this here.

pandafy (changes requested):

openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
```diff
@@ -72,6 +82,12 @@ def mocked_exec_upgrade_success(command, exit_codes=None, timeout=None):
         raise CommandFailedException()
 
 
+def mocked_exec_uuid_mismatch(command, exit_codes=None, timeout=None):
+    if command == 'uci get openwisp.http.uuid':
+        return ['different-uuid', 0]
```
nemesifier, line 87: Place a real UUID here please, for making the test more realistic.

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,38 @@ def upload(self, image_file, remote_path):
...
+            'uci get openwisp.http.uuid', exit_codes=[0, 1]
+        )
+        if exit_code == 1 or not output.strip():
+            self.log(_('Could not read device UUID from configuration'))
```
nemesifier, line 169: I haven't seen any assertion verifying for the presence of this log line, did you add any?.

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,38 @@ def upload(self, image_file, remote_path):
...
+            raise UpgradeAborted(
+                'Could not verify device identity: UUID not found in device configuration'
```
nemesifier, line 171: I looked around the code base and I've noticed we always just call `raise UpgradeAborted()`, without any message, so I believe there's no need to specify an exception message here because we are already providing an error message by calling the `self.log()` method (it's redundant but with no benefits).

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,38 @@ def upload(self, image_file, remote_path):
...
+            )
+        device_config_uuid = output.strip()
+        if str(device_uuid) != device_config_uuid:
+            self.log(
+                _(
+                    'Device UUID mismatch: expected {expected}, '
+                    'found {found} in device configuration'
```
nemesifier, line 178: Are you testing for the presence of this in the tests?

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,38 @@ def upload(self, image_file, remote_path):
...
+                ).format(expected=device_uuid, found=device_config_uuid)
+            )
+            raise UpgradeAborted(
+                'Upgrade aborted: device UUID mismatch detected, '
+                'this could indicate an attempt to upgrade the wrong device'
```
nemesifier, line 183: I looked around the code base and I've noticed we always just call `raise UpgradeAborted()`, without any message, so I believe there's no need to specify an exception message here because we are already providing an error message by calling the `self.log()` method (it's redundant but with no benefits).

nemesifier (changes requested):
Good progress @youhaveme9, see my comments below.

openwisp_firmware_upgrader/upgraders/openwrt.py
```diff
@@ -155,8 +155,33 @@ def upload(self, image_file, remote_path):
...
+        device_config_uuid = output.strip()
+        if str(device_uuid) != device_config_uuid:
+            self.log(
+                _(
+                    'Device UUID mismatch: expected {expected}, '
+                    'found {found} in device configuration'
+                ).format(expected=device_uuid, found=device_config_uuid)
+            )
+            raise UpgradeAborted()
+        self.log(_('Device identity verified successfully'))
```
nemesifier, line 180: Add an assertion for the presence of this too, I had forgot to ask you this. Should be quick though.
In general, we should always assert for the presence of important UI elements we are implementing, to protect us from unwanted changes that could remove these.

nemesifier (changes requested):

Aryamanz29 (commented):
LGTM (great to see proper log `assertions` in place now) - just a heads-up, it seems the selenium test is failing for `Python 3.9` with `Django 4.2.0` based on the build logs:

https://github.com/openwisp/openwisp-firmware-upgrader/actions/runs/14605007340/job/40971977302?pr=299
```
ERROR: test_upgrader_with_unsupported_upgrade_options (openwisp_firmware_upgrader.tests.test_selenium.TestDeviceAdmin) [Test BatchUpgradeOperation]
```

nemesifier (approved):
The flaky test is unrelated to this PR and @youhaveme9 has opened an issue for it https://github.com/openwisp/openwisp-firmware-upgrader/issues/304, we also have https://github.com/openwisp/openwisp-utils/issues/464 on our list, which should help.

PR #305, opened 2025-04-23
[fix] Firmware image files are not deleted when deleting build

### LINKED ISSUE #301: [bug] Firmware image files are not deleted when deleting build, category or organization
**Describe the bug**

When deleting a build or a category, the delete operation is cascaded to related models.

Unfortunately due to how Django works, the method `AbstractFirmwareImage.delete()` is not called in this case, [quoting from the Django docs](https://docs.djangoproject.com/en/5.2/topics/db/models/#overriding-predefined-model-methods):

> Overridden model methods are not called on bulk operations
> 
> Note that the [delete()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.delete) method for an object is not necessarily called when [deleting objects in bulk using a QuerySet](https://docs.djangoproject.com/en/5.2/topics/db/queries/#topics-db-queries-delete) or as a result of a [cascading delete](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.ForeignKey.on_delete). To ensure customized delete logic gets executed, you can use [pre_delete](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.pre_delete) and/or [post_delete](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.post_delete) signals.
> 
> Unfortunately, there isn’t a workaround when [creating](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.bulk_create) or [updating](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.update) objects in bulk, since none of [save()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.save), [pre_save](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.pre_save), and [post_save](https://docs.djangoproject.com/en/5.2/ref/signals/#django.db.models.signals.post_save) are called.

This is probably causing a lot of wasted storage space.

**Steps To Reproduce**
1. Create a build and upload a firmware image (any file will work)
2. Verify the filesystem path of the uploaded file
3. Delete the build
4. Verify that the firmware image file was not deleted

The same happens when deleting a category and I guess it also happens if an Organization is deleted.

**Expected behavior**
The firmware image files should be removed.

**Proposed solution**
I think we could use `pre_delete` receivers listening to `Build`, `Category` and `Organization`.
From what I see here:
https://github.com/django/django/blob/89abecc75d7eadab681f29be5237972c6c2f997b/django/db/models/deletion.py#L382-L417
It seems that all the delete operations are wrapped in a DB transaction.
Therefore I think we can do something like the following:

- In the post_delete receiver, collect a list of filesystem paths that have to be deleted
- Use transaction.on_commit to launch a new celery task in the background to which we should pass the list of filesystem paths to delete (we're doing the same in different openwisp modules)
- The celery task will be executed once the transaction was completed succesfully and we can iterate on the list of filesystem paths to perform the deletion in the background

youhaveme9:
Firmware image files are not deleted when deleting build

## Checklist

Closes #301

### DIFF
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -161,9 +161,11 @@ class AbstractBuild(TimeStampedEditableModel):
         )
         batch.full_clean()
         batch.save()
-        transaction.on_commit(
-            lambda: batch_upgrade_operation.delay(batch.pk, firmwareless)
-        )
+
+        def schedule_batch_upgrade():
+            batch_upgrade_operation.delay(batch.pk, firmwareless)
+
+        transaction.on_commit(schedule_batch_upgrade)
         return batch
 
     def _find_related_device_firmwares(self, select_devices=False):
@@ -250,11 +252,23 @@ class AbstractFirmwareImage(TimeStampedEditableModel):
             raise ValidationError({'type': 'Could not find boards for this type'})
 
     def delete(self, *args, **kwargs):
-        """
-        Delete only deletes the database record
-        The actual file deletion is handled by the signal handlers
-        """
         super().delete(*args, **kwargs)
+        self._remove_file()
+
+    def _remove_file(self):
+        firmware_filename = self.file.name
+        self.file.storage.delete(firmware_filename)
+        # Delete the file directory
+        dir = os.path.split(firmware_filename)[0]
+        if dir:
+            try:
+                self.file.storage.delete(dir)
+            except OSError as error:
+                # A build can have multiple firmware images,
+                # therefore, deleting directory might not be
+                # always feasible.
+                if 'Directory not empty' not in str(error):
+                    raise error
 
     def _clean_type(self):
         """
@@ -337,9 +351,13 @@ class AbstractDeviceFirmware(TimeStampedEditableModel):
             operation.batch = batch
         operation.full_clean()
         operation.save()
+
         # launch ``upgrade_firmware`` in the background (celery)
         # once changes are committed to the database
-        transaction.on_commit(lambda: upgrade_firmware.delay(operation.pk))
+        def schedule_upgrade():
+            upgrade_firmware.delay(operation.pk)
+
+        transaction.on_commit(schedule_upgrade)
         return operation
 
     @classmethod
@@ -387,14 +405,19 @@ class AbstractDeviceFirmware(TimeStampedEditableModel):
         if instance.device.model not in REVERSE_FIRMWARE_IMAGE_MAP:
             return
 
-        transaction.on_commit(lambda: create_device_firmware.delay(instance.device.pk))
+        def schedule_device_firmware():
+            create_device_firmware.delay(instance.device.pk)
+
+        transaction.on_commit(schedule_device_firmware)
 
     @classmethod
     def auto_create_device_firmwares(cls, instance, created, **kwargs):
         if created:
-            transaction.on_commit(
-                lambda: create_all_device_firmwares.delay(instance.pk)
-            )
+
+            def schedule_all_device_firmwares():
+                create_all_device_firmwares.delay(instance.pk)
+
+            transaction.on_commit(schedule_all_device_firmwares)
 
     @classmethod
     def get_image_queryset_for_device(cls, device, device_firmware=None):
--- a/openwisp_firmware_upgrader/base/models.py
+++ b/openwisp_firmware_upgrader/base/models.py
@@ -250,23 +250,11 @@ class AbstractFirmwareImage(TimeStampedEditableModel):
             raise ValidationError({'type': 'Could not find boards for this type'})
 
     def delete(self, *args, **kwargs):
+        """
+        Delete only deletes the database record
+        The actual file deletion is handled by the signal handlers
+        """
         super().delete(*args, **kwargs)
-        self._remove_file()
-
-    def _remove_file(self):
-        firmware_filename = self.file.name
-        self.file.storage.delete(firmware_filename)
-        # Delete the file directory
-        dir = os.path.split(firmware_filename)[0]
-        if dir:
-            try:
-                self.file.storage.delete(dir)
-            except OSError as error:
-                # A build can have multiple firmware images,
-                # therefore, deleting directory might not be
-                # always feasible.
-                if 'Directory not empty' not in str(error):
-                    raise error
 
     def _clean_type(self):
         """
--- a/openwisp_firmware_upgrader/signals.py
+++ b/openwisp_firmware_upgrader/signals.py
@@ -30,7 +30,11 @@ def delete_build_files(sender, instance, **kwargs):
     logger.info(f"Collected file paths: {paths}")
     if paths:
         logger.info("Scheduling delete_firmware_files task")
-        transaction.on_commit(lambda: delete_firmware_files.delay(paths))
+
+        def schedule_delete_task():
+            delete_firmware_files.delay(paths)
+
+        transaction.on_commit(schedule_delete_task)
 
 
 def delete_category_files(sender, instance, **kwargs):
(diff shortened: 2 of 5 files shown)


purhan (commented):
Thanks, looks good on a high level, I'll test and let you know. Looks like the checks are failing due to formatting.

openwisp_firmware_upgrader/base/models.py
```diff
@@ -250,23 +250,11 @@ def clean(self):
...
 
     def delete(self, *args, **kwargs):
+        """
+        Delete only deletes the database record
+        The actual file deletion is handled by the signal handlers
+        """
         super().delete(*args, **kwargs)
-        self._remove_file()
-
-    def _remove_file(self):
```
nemesifier, line 256: Why are you removing this?

youhaveme9: The actual file deletion is handled by the signal handlers
The FirmwareImage model's delete method is calling _remove_file() after super().delete(), which means it's trying to delete files that the signal handlers are also trying to delete

nemesifier: The `delete()` method gets called when an instance of FirmwareImage gets deleted.
However, the delete method is not called when the deletion is cascaded from parent models.

I think this is not right. Are you sure this is right?

youhaveme9:
I have implemented the solution to handle firmware file deletion when Build, Category, or Organization models are deleted. The implementation:

1. Uses Django signals to detect when models are deleted
2. Collects all related firmware file paths that need to be deleted
3. Uses transaction.on_commit with named functions to ensure the deletion task runs after the transaction is complete
4. Executes file deletion in a background celery task

openwisp_firmware_upgrader/signals.py
pandafy, line 1: The signal file in Django is used to to include signal definitions, e.g. post_save, post_delete, etc. 

I recommend that you convert these functions into classmethod and add them to AbstractFirmwareImage class. 

- `organization_pre_delete_handler`
- `category_pre_delete_handler`
- `build_pre_delete_handler`

You can look into other OpenWISP modules (e.g. controller) to check how we attach signal receivers in OpenWISP. Let me know if you have any doubts.

openwisp_firmware_upgrader/signals.py
```diff
@@ -0,0 +1,85 @@
...
+    logger.info(f"Build delete signal received for build {instance.pk}")
+    firmware_images = instance.firmwareimage_set.all()
+    logger.info(f"Found {firmware_images.count()} firmware images to delete")
+    paths = _collect_firmware_files(firmware_images)
+    logger.info(f"Collected file paths: {paths}")
+    if paths:
+        logger.info("Scheduling delete_firmware_files task")
+
+        def schedule_delete_task():
+            delete_firmware_files.delay(paths)
```
pandafy, line 34: We don't need to define a new function here. You can use Python lambda function with `transaciton.on_commit`.

openwisp_firmware_upgrader/signals.py
```diff
@@ -0,0 +1,85 @@
...
+            paths.append(image.file.name)
+    return paths
+
+
+def delete_build_files(sender, instance, **kwargs):
+    """
+    Delete firmware files when a Build is deleted
+    """
+    logger.info(f"Build delete signal received for build {instance.pk}")
+    firmware_images = instance.firmwareimage_set.all()
```
pandafy, line 26: `all()` evaluates the queryset immediately. It is generally preferred to defer query evaluation to as late in code as possible. 

We can further optimize this queryset by only retrieving the file field from the database. 

```
instance.firmwareimage_set.values_list('file', flat=True)
```

This directly returns a list of file paths.

openwisp_firmware_upgrader/signals.py
```diff
@@ -0,0 +1,85 @@
...
+    logger.info(f"Found {firmware_images.count()} firmware images to delete")
```
pandafy, line 27: This is making additional query to the database. We can get this value by calling `len()` on `firmware_images`.

openwisp_firmware_upgrader/signals.py
```diff
@@ -0,0 +1,85 @@
...
+    Delete firmware files when a Category is deleted
+    """
+    logger.info(f"Category delete signal received for category {instance.pk}")
+    firmware_images = []
+    builds = instance.build_set.all()
+    logger.info(f"Found {builds.count()} builds to process")
+    for build in builds:
+        images = build.firmwareimage_set.all()
+        logger.info(f"Found {images.count()} firmware images in build {build.pk}")
+        firmware_images.extend(images)
```
pandafy, line 50: Similarly, we can optimize this by using the following query

```
FirmwareImage.objects.filter(build__category_id=category.id).values_list('file', flat=True)
```

Pay attention, that we are creating a join at the database level which is more efficient than looping over all the objects.

openwisp_firmware_upgrader/signals.py
```diff
@@ -0,0 +1,85 @@
...
+    firmware_images = []
+    categories = instance.category_set.all()
+    logger.info(f"Found {categories.count()} categories to process")
+    for category in categories:
+        builds = category.build_set.all()
+        logger.info(f"Found {builds.count()} builds in category {category.pk}")
+        for build in builds:
+            images = build.firmwareimage_set.all()
+            logger.info(f"Found {images.count()} firmware images in build {build.pk}")
+            firmware_images.extend(images)
```
pandafy, line 76: Optimize this similarly using database join.

openwisp_firmware_upgrader/tasks.py
```diff
@@ -15,6 +16,60 @@
 logger = logging.getLogger(__name__)
 
 
+@shared_task(base=OpenwispCeleryTask, bind=True)
+def delete_firmware_files(self, file_paths):
+    """
+    Deletes firmware files from storage in the background
+    """
+    logger.info(f"Starting delete_firmware_files task with paths: {file_paths}")
```
pandafy, line 24: I think the logging here is too verbose. Either remove the obvious log lines or change the log level to DEBUG as it will create unnecessary noise.

openwisp_firmware_upgrader/tasks.py
```diff
@@ -15,6 +16,60 @@
...
+    logger.info(f"Storage location: {storage.location}")
+
+    for path in file_paths:
+        logger.info(f"Attempting to delete file: {path}")
+        try:
+            try:
+                exists = storage.exists(path)
+                logger.info(f"File exists: {exists}")
+                if exists:
+                    full_path = storage.path(path)
```
pandafy, line 39: What happens if you don't build the full path? What if you just delete without check if the file exists? Does it raise an exception?

openwisp_firmware_upgrader/tasks.py
```diff
@@ -15,6 +16,60 @@
...
+                dir_path = '/'.join(path.split('/')[:-1])
+                if dir_path:
+                    logger.info(f"Attempting to delete directory: {dir_path}")
+                    try:
+                        if storage.exists(dir_path):
+                            full_dir_path = storage.path(dir_path)
+                            logger.info(f"Full directory path: {full_dir_path}")
+                            if not os.listdir(
+                                full_dir_path
+                            ):  # Check if directory is empty
```
pandafy, line 56: Don't use the `os` module here. Use the method of the [storage class](https://docs.djangoproject.com/en/5.1/ref/files/storage/#the-storage-class). The storage could be filesystem storage or an S3 bucket.

openwisp_firmware_upgrader/tests/test_signals.py
```diff
@@ -0,0 +1,141 @@
...
+        self.assertTrue(os.path.exists(file_path))
+
+        with patch(
+            'openwisp_firmware_upgrader.tasks.delete_firmware_files.delay'
+        ) as mock:
+            build.delete()
+            mock.assert_called_once()
+            file_paths = mock.call_args[0][0]
+            self.assertEqual(len(file_paths), 1)
+            self.assertEqual(file_paths[0], fw.file.name)
```
pandafy, line 58: Let's not mock these. Instead, we should verify that deleting the `build` deletes these files.

pandafy (changes requested):

PR #306, opened 2025-04-25
[change] Improve endpoints to download firmware images 

### LINKED ISSUE #69: [change] Improve endpoints to download firmware images
Right now, the API endpoint which shows the image information, has a field, called "file", which links to the private-storage view.
But we also have an API endpoint to download the firmware, although it's implemented differently than the private storage view.

It also looks that the private storage view does not check if the user has "view permission" on the object. So it would be theoretically possible that a staff user without premissions to view firmware images can still download the firmware image, which is a problem.

Therefore, I think we should do the following things:

- [x] Find a way to call `openwisp_firmware_upgrader.private_storage.views.FirmwareImageDownloadView` from `openwisp_firmware_upgrader.api.views.FirmwareImageDownloadView`, to avoid duplicating logic
- [ ] Change the `file` attribute of `FirmwareImageSerializer` to point to the API URL (because API supports token authentication (we need a test that ensures the new URL is not the one generated by private_storage)
- [ ]  Ensure the private storage view checks whether the user has the view permission on FirmwareImage objects (we need a test for this), once this is done, we can try to set permission_classes to an empty list on the API FirmwareImageDownloadView, to avoid checking permissions twice, in theory it should work

youhaveme9:
## Checklist

- N/A I have updated the documentation.

Closes #69

### DIFF
--- a/openwisp_firmware_upgrader/api/serializers.py
+++ b/openwisp_firmware_upgrader/api/serializers.py
@@ -46,13 +46,14 @@ class CategoryRelationSerializer(BaseSerializer):
 
 
 class FirmwareImageSerializer(BaseSerializer):
-    file = serializers.SerializerMethodField()
+    file = serializers.FileField(required=False)
+    file_url = serializers.SerializerMethodField(read_only=True)
 
     def validate(self, data):
         data["build"] = self.context["view"].get_parent_queryset().get()
         return super().validate(data)
 
-    def get_file(self, obj):
+    def get_file_url(self, obj):
         request = self.context.get('request')
         if request and getattr(obj, 'pk', None):
             return request.build_absolute_uri(
@@ -62,6 +63,13 @@ class FirmwareImageSerializer(BaseSerializer):
                 )
             )
         return obj.file.url if hasattr(obj, 'file') else None
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        # Replace 'file' with the URL from file_url
+        if 'file_url' in ret:
+            ret['file'] = ret.pop('file_url')
+        return ret
 
     class Meta(BaseMeta):
         model = FirmwareImage
--- a/openwisp_firmware_upgrader/api/serializers.py
+++ b/openwisp_firmware_upgrader/api/serializers.py
@@ -1,5 +1,6 @@
 import swapper
 from django.core.exceptions import ValidationError
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 from rest_framework import serializers
 
@@ -45,9 +46,22 @@ class CategoryRelationSerializer(BaseSerializer):
 
 
 class FirmwareImageSerializer(BaseSerializer):
+    file = serializers.SerializerMethodField()
+
     def validate(self, data):
         data["build"] = self.context["view"].get_parent_queryset().get()
         return super().validate(data)
+
+    def get_file(self, obj):
+        request = self.context.get('request')
+        if request and getattr(obj, 'pk', None):
+            return request.build_absolute_uri(
+                reverse(
+                    'upgrader:api_firmware_download',
+                    args=[obj.build.pk, obj.pk],
+                )
+            )
+        return obj.file.url if hasattr(obj, 'file') else None
 
     class Meta(BaseMeta):
         model = FirmwareImage
--- a/openwisp_firmware_upgrader/api/serializers.py
+++ b/openwisp_firmware_upgrader/api/serializers.py
@@ -52,17 +52,17 @@ class FirmwareImageSerializer(BaseSerializer):
 
     def to_representation(self, instance):
         ret = super().to_representation(instance)
-        request = self.context.get('request')
-        if request and getattr(instance, 'pk', None):
-            ret['file'] = request.build_absolute_uri(
+        request = self.context.get("request")
+        if request and getattr(instance, "pk", None):
+            ret["file"] = request.build_absolute_uri(
                 reverse(
-                    'upgrader:api_firmware_download',
+                    "upgrader:api_firmware_download",
                     args=[instance.build.pk, instance.pk],
                 )
             )
-        elif hasattr(instance, 'file'):
-            ret['file'] = reverse(
-                'upgrader:api_firmware_download',
+        elif hasattr(instance, "file"):
+            ret["file"] = reverse(
+                "upgrader:api_firmware_download",
                 args=[instance.build.pk, instance.pk],
             )
         return ret
--- a/openwisp_firmware_upgrader/api/serializers.py
+++ b/openwisp_firmware_upgrader/api/serializers.py
@@ -47,28 +47,23 @@ class CategoryRelationSerializer(BaseSerializer):
 
 class FirmwareImageSerializer(BaseSerializer):
     file = serializers.FileField(required=False)
-    file_url = serializers.SerializerMethodField(read_only=True)
 
     def validate(self, data):
         data["build"] = self.context["view"].get_parent_queryset().get()
         return super().validate(data)
 
-    def get_file_url(self, obj):
-        request = self.context.get('request')
-        if request and getattr(obj, 'pk', None):
-            return request.build_absolute_uri(
-                reverse(
-                    'upgrader:api_firmware_download',
-                    args=[obj.build.pk, obj.pk],
-                )
-            )
-        return obj.file.url if hasattr(obj, 'file') else None
-
     def to_representation(self, instance):
         ret = super().to_representation(instance)
-        # Replace 'file' with the URL from file_url
-        if 'file_url' in ret:
-            ret['file'] = ret.pop('file_url')
+        request = self.context.get('request')
+        if request and getattr(instance, 'pk', None):
+            ret['file'] = request.build_absolute_uri(
+                reverse(
+                    'upgrader:api_firmware_download',
+                    args=[instance.build.pk, instance.pk],
+                )
+            )
+        elif hasattr(instance, 'file'):
+            ret['file'] = instance.file.url
         return ret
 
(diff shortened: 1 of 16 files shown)


openwisp_firmware_upgrader/private_storage/views.py
```diff
@@ -12,9 +13,26 @@ class FirmwareImageDownloadView(PrivateStorageDetailView):
...
+            return False
+
+        # For authorized users, check view permission
+        # Only if they're not superusers (superusers have all permissions)
+        if not user.is_superuser:
+            perm_codename = get_permission_codename('view', self.model._meta)
+            view_perm = f'{self.model._meta.app_label}.{perm_codename}'
+            has_view_perm = user.has_perm(view_perm, self.object)
+            if not has_view_perm:
+                return False
```
pandafy, line 33: Did you consider using the [PermissionRequiredMixin](https://docs.djangoproject.com/en/5.2/topics/auth/default/#the-permissionrequiredmixin-mixin) instead of performing these checks manually?

openwisp_firmware_upgrader/api/serializers.py
```diff
@@ -45,10 +46,31 @@ class Meta:
...
+                )
+            )
+        return obj.file.url if hasattr(obj, 'file') else None
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        # Replace 'file' with the URL from file_url
+        if 'file_url' in ret:
+            ret['file'] = ret.pop('file_url')
+        return ret
```
pandafy, line 72: Since we are already overriding the value of `field` value in the to_representation, then why do we need to add another field `file_url` to the serializer? We can directly call the `get_file_url`

pandafy (changes requested):
Looks on the correct track. Check my suggestion below.

openwisp_firmware_upgrader/tests/test_api.py
```diff
@@ -970,9 +985,11 @@ def test_firmware_download(self):
         url = reverse('upgrader:api_firmware_download', args=[image.build.pk, image.pk])
         with self.subTest("Test as operator"):
```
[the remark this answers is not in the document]
youhaveme9: we are testing it for admin too below

openwisp_firmware_upgrader/tests/test_private_storage.py
```diff
@@ -68,9 +74,41 @@ def test_staff_operator_with_same_organization(self):
             user=staff_user, organization=self.test_org, is_admin=True
         )
         self.client.force_login(staff_user)
-        self._download_firmware_assert_status(200)
+        self._download_firmware_assert_status(403)
```
pandafy, line 77: Why did you change this? Aren't operators allowed to view the FirmwareImage? 

https://github.com/openwisp/openwisp-firmware-upgrader/blob/ee878d5f080059b0b47ca3246c4fa207c043ad43/openwisp_firmware_upgrader/migrations/__init__.py#L28-L34

openwisp_firmware_upgrader/tests/test_private_storage.py
```diff
@@ -68,9 +74,41 @@ def test_staff_operator_with_same_organization(self):
...
 
     def test_superuser(self):
         user = self._get_admin()
         self.client.force_login(user)
         self._download_firmware_assert_status(200)
+
+    def test_view_permission_check(self):
```
pandafy, line 84: Are you sure this test is correct? I believe the code comments are outdated. 

A user being **Organization Admin** and a user having **Administrator** group are two different things. Is this clear?

openwisp_firmware_upgrader/tests/test_api.py
```diff
@@ -817,23 +821,34 @@ def test_firmware_unauthorized(self):
...
         url = reverse('upgrader:api_firmware_download', args=[image.build.pk, image.pk])
         with self.subTest(url=url):
             with self.assertNumQueries(1):
                 r = client.get(url)
-            self.assertEqual(r.status_code, 401)
+            self.assertEqual(r.status_code, 403)
 
     def test_firmware_list(self):
         image = self._create_firmware_image()
-        self._create_firmware_image(type=self.TPLINK_4300_IL_IMAGE)
```
pandafy, line 828: Why did we remove this?

pandafy (changes requested):
I am not convinced with the changes to the tests. Can you clarify my doubts?

openwisp_firmware_upgrader/tests/test_api.py
youhaveme9: I was debugging something
I missed to restore it before pushing the code
my bad

openwisp_firmware_upgrader/tests/test_private_storage.py
youhaveme9: Yes clear now

openwisp_firmware_upgrader/private_storage/views.py
```diff
@@ -1,19 +1,44 @@
...
+        # Get object first since it's needed for organization check
+        self.object = self.get_object()
+
+        # Superusers can always access
+        if user.is_superuser:
+            return True
+
+        return user.is_staff and user.is_manager(
+            self.object.build.category.organization
         )
```
pandafy, line 42: Does the logic stays same as before?

openwisp_firmware_upgrader/tests/test_private_storage.py
```diff
@@ -74,3 +80,44 @@ def test_superuser(self):
...
+        org_user.save()
+
+        # No access (missing staff status)
+        self._download_firmware_assert_status(403)
+
+        # Remove view permission
+        staff_user.user_permissions.remove(view_perm)
+
+        # No access (missing view permission)
+        self._download_firmware_assert_status(403)
```
pandafy, line 123: Looks good, can you refactor the code to use `subTest`? It helps to organize code for testing different scenarios. You can look around the code for reference. 

https://github.com/search?q=repo%3Aopenwisp%2Fopenwisp-firmware-upgrader+subTest&type=code

pandafy (changes requested):

Aryamanz29 (approved):
LGTM

openwisp_firmware_upgrader/tests/test_private_storage.py
```diff
@@ -74,3 +80,47 @@ def test_superuser(self):
...
         self.client.force_login(user)
         self._download_firmware_assert_status(200)
+
+    def test_view_permission_check(self):
+        staff_user = self._get_operator()
+        self.client.force_login(staff_user)
+        org = self.image.build.category.organization
+
+        with self.subTest("Test initial access without permissions"):
+            self._download_firmware_assert_status(403)
```
pandafy, line 90: This test is incorrect (or, the description is misleading). 

The `Operator` group has read-only permission for the FirmwareImage model 

https://github.com/openwisp/openwisp-firmware-upgrader/blob/e47bdcf6edb7d2d5db7a0ce0c7198b3dceb60f03/openwisp_firmware_upgrader/migrations/__init__.py#L28-L34

So, this code only tests that the user is not an org admin.

pandafy (changes requested):
@youhaveme9 I am still not convinced with the changes to the test-suite. I expect to test the following scenarios from both REST API and the private storage view 

| #   | Access the view with                                           | is staff | is org admin | has view permission |
|-----|----------------------------------------------------------------|----------|---------------|----------------------|
| 1.  | a user without any permission                                  | ❌       | ❌            | ❌                   |
| 2.  | staff user                                                     | ✅       | ❌            | ❌                   |
| 3.  | staff user who is org admin of a different organization        | ✅       | ✅            | ✅                   |
| 4.  | staff user who is org admin of same organization               | ✅       | ✅            | ✅                   |


for 3 and 4, you can add the user in "Operator" or "Administrator" group. 

You can create a re-usable helper functions if you prefer.

openwisp_firmware_upgrader/tests/test_private_storage.py
```diff
@@ -74,3 +80,47 @@ def test_superuser(self):
...
+        org_user.is_admin = False
+        org_user.save()
+
+        # Remove staff status
+        staff_user.is_staff = False
+        staff_user.save()
+
+        # Restore org manager status
+        org_user.is_admin = True
+        org_user.save()
```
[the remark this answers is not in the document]
youhaveme9: My bad 
I somehow missed the assertion after I removed the admin status from the user
I didn't see before committing the code :\

pandafy (dismissed):
LGTM! ✅ 

In order to move quickly, I went ahead and improved the structure of the tests. Now, the private storage view and API view shares the same test mixin. "

@youhaveme9 I ask you to please review my changes in https://github.com/openwisp/openwisp-firmware-upgrader/pull/306/commits/0384d62c03d72dc385a0414fc0f7c64ae7251aa2 and let me know if I missed any test scenario. 

If everything is good, please do a final rebase on the latest master.

pandafy (dismissed):

nemesifier (changes requested):
The 403 forbidden result when user doesn't have permissions on the firmware model is weird:

Screenshot from 2025-08-16 15-25-36

I believe this is due to the fact that we're using `django.contrib.auth.mixins.PermissionRequiredMixin`, which is not designed for Django REST Framework Views.

DRF has its own thing, `DjangoModelPermissions` which we have extended in openwisp-utils, please read:

- https://openwisp.io/docs/dev/users/developer/django-rest-framework-utils.html#djangomodelpermissions
- https://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions

I think we should use `openwisp_users.api.permissions.DjangoModelPermissions` in this PR, the result would be more consistent with the rest of the OW codebase and we should end up

nemesifier (approved):

PR #322, opened 2025-05-28
[fix] Ensure firmware image files get deleted from filesystem

### LINKED ISSUE #301: [bug] Firmware image files are not deleted when deleting build, category or organization

youhaveme9:
Firmware image files were not deleted when deleting build, category or organization. This patch fixes #301.

## Checklist

- N/A I have updated the documentation.

Closes #301

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_utils.py
+++ b/openwisp_firmware_upgrader/tests/test_utils.py
@@ -3,11 +3,8 @@ from unittest.mock import patch
 from django.test import TestCase
 
 from .. import settings as app_settings
-from ..swapper import load_model
 from ..utils import get_upgrader_class_from_device_connection
 from .base import TestUpgraderMixin
-
-FirmwareImage = load_model("FirmwareImage")
 
 
 class TestUtils(TestUpgraderMixin, TestCase):


openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -12,7 +12,7 @@
 from .. import settings as app_settings
 from ..hardware import FIRMWARE_IMAGE_MAP, REVERSE_FIRMWARE_IMAGE_MAP
 from ..swapper import load_model
-from ..tasks import upgrade_firmware
+from ..tasks import delete_firmware_files, upgrade_firmware
```
nemesifier, line 15: I am not in favour of the way these tests are written.

Tests for bugs should fail without the fix, that is, if I run these tests on master they should fail (all assertions/subtests), while they should pass with the patch.

However, if you try to run these tests on master they can't be run because `delete_firmware_files` is not yet available. That would be ok if we were adding test for a feature, but we need tests that replicate the bug.

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -310,6 +310,75 @@ def test_upgrade_operation_retention_on_image_delete(self):
...
+        file_storage_backend = FirmwareImage.file.field.storage
+        build = self._create_build()
+        image = self._create_firmware_image(build=build)
+        file_name = image.file.name
+
+        # Delete the build
+        build.delete()
+
+        # Execute the Celery task synchronously
+        delete_firmware_files([file_name])
```
nemesifier, line 342: Why is calling this explicitly needed? Isn't this supposed to be called automatically due to `CELERY_TASK_ALWAYS_EAGER = True`?

nemesifier (changes requested):

openwisp_firmware_upgrader/base/models.py
```diff
@@ -278,6 +279,57 @@ def _clean_type(self):
...
+        files_to_delete = []
+
+        # Get all firmware images related to builds in categories of this organization
+        for image in cls.objects.filter(build__category__organization=instance):
+            if image.file:
+                files_to_delete.append(image.file.name)
+
+        # Schedule file deletion after transaction is committed
+        if files_to_delete:
+            transaction.on_commit(lambda: delete_firmware_files.delay(files_to_delete))
```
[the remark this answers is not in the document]
Aryamanz29: let's keep it DRY :wink:

openwisp_firmware_upgrader/tasks.py
```diff
@@ -83,3 +83,27 @@ def create_all_device_firmwares(self, firmware_image_id):
     queryset = Device.objects.filter(os=fw_image.build.os)
     for device in queryset.iterator():
         DeviceFirmware.create_for_device(device, fw_image)
+
+
+@shared_task(base=OpenwispCeleryTask)
+def delete_firmware_files(files_to_delete):
+    FirmwareImage = load_model("FirmwareImage")
+    storage = FirmwareImage.file.field.storage
+
```
Aryamanz29, line 92: I'd add a docstring here, something like:

```py
   """
    Celery task to delete firmware image files and their parent directories if empty.

    Args:
        files_to_delete (list[str]): A list of file paths (relative to the storage backend)
                                     that should be deleted.

    Behavior:
        - Deletes each firmware file using the storage backend associated with FirmwareImage.file.
        - After deleting a file, checks its parent directory:
            - If the directory is empty, it is deleted.
            - If the directory is not empty or deletion fails, logs a debug/warning message.
        - All exceptions during file or directory deletion are logged,
          but do not interrupt the task's execution for other files.
    """
```

openwisp_firmware_upgrader/tasks.py
```diff
@@ -83,3 +83,27 @@ def create_all_device_firmwares(self, firmware_image_id):
...
+    for file_path in files_to_delete:
+        try:
+            storage.delete(file_path)
+            dir_path = "/".join(file_path.split("/")[:-1])
+            if dir_path:
+                try:
+                    dirs, files = storage.listdir(dir_path)
+                    if not dirs and not files:
+                        storage.delete(dir_path)
```
Aryamanz29, line 101: I would also add logger statement here ^^

```py
logger.info("Deleted empty directory: %s", dir_path)
```

openwisp_firmware_upgrader/tasks.py
```diff
@@ -83,3 +83,27 @@ def create_all_device_firmwares(self, firmware_image_id):
...
+                    else:
+                        logger.debug(
+                            f"Directory {dir_path} is not empty, skipping deletion"
+                        )
+                except Exception as error:
+                    logger.debug(f"Could not delete directory {dir_path}: {str(error)}")
```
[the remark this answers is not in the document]
Aryamanz29: Yup, using `warning` or `error` here seems more appropriate than debug - especially since it's something that might indicate a problem during cleanup 👍

openwisp_firmware_upgrader/tasks.py
```diff
@@ -83,3 +83,27 @@ def create_all_device_firmwares(self, firmware_image_id):
...
+            storage.delete(file_path)
+            dir_path = "/".join(file_path.split("/")[:-1])
+            if dir_path:
+                try:
+                    dirs, files = storage.listdir(dir_path)
+                    if not dirs and not files:
+                        storage.delete(dir_path)
+                    else:
+                        logger.debug(
+                            f"Directory {dir_path} is not empty, skipping deletion"
```
Aryamanz29, line 104: Let’s use `%-style` formatting with `logger.*()` since the Python logging library is optimized for it - it defers string interpolation until it’s actually needed (based on the log level), which is more efficient. 
https://stackoverflow.com/questions/54367975/python-3-7-logging-f-strings-vs

Aryamanz29: ex:
```py
logger.debug("Directory %s is not empty, skipping deletion", dir_path)
```

openwisp_firmware_upgrader/tests/test_models.py
```diff
@@ -492,3 +510,39 @@ def test_delete_firmware_image_file(self):
...
+        file_storage_backend = FirmwareImage.file.field.storage
+        org = self._get_org()
+        category = self._create_category(organization=org)
+        build = self._create_build(category=category)
+        image = self._create_firmware_image(build=build)
+        file_name = image.file.name
+        # Delete the organization
+        org.delete()
+        # Check that the file was deleted
+        self.assertEqual(file_storage_backend.exists(file_name), False)
```
Aryamanz29, line 548: Good to see that we have tests covering the happy paths - when the user deletes a build, category, or organization. ✅

However, we should also add tests for other scenarios to fully cover the task's behavior, as @purhan mentioned:

* Cases where deletion fails (e.g. storage backend raises an exception)
* When the directory (`dir_path`) is not empty
* Any other exceptions we might catch while deleting files or directories

Also, don't forget to **assert the log messages** we emit in those scenarios - it's important to verify observability behavior too. 🙏

openwisp_firmware_upgrader/base/models.py
```diff
@@ -162,7 +164,7 @@ def batch_upgrade(self, firmwareless, upgrade_options=None):
         batch.full_clean()
         batch.save()
         transaction.on_commit(
-            lambda: batch_upgrade_operation.delay(batch.pk, firmwareless)
+            partial(batch_upgrade_operation.delay, batch.pk, firmwareless)
```
nemesifier, line 167: I just would like to understand if there was a need to change this or if you just wanted to rewrite it?
If there was a need, what was it?

openwisp_firmware_upgrader/utils.py
```diff
@@ -41,3 +44,79 @@ def get_upgrader_class_from_device_connection(device_conn):
...
+        - Collects file paths from images that have files
+        - Schedules asynchronous deletion of collected files after transaction commit
+        - Uses the delete_firmware_files Celery task for actual deletion
+    """
+    from .tasks import delete_firmware_files  # Import here to avoid circular imports
+
+    files_to_delete = []
+
+    # Get all firmware images matching the filter criteria
+    for image in firmware_image_class.objects.filter(**filter_kwargs):
```
nemesifier, line 116:
```suggestion
    for image in firmware_image_class.objects.iterator(**filter_kwargs):
```

See https://docs.djangoproject.com/en/5.2/ref/models/querysets/#iterator.

openwisp_firmware_upgrader/base/models.py
```diff
@@ -278,6 +279,57 @@ def _clean_type(self):
...
+        files_to_delete = []
+
+        # Get all firmware images related to this build
+        for image in cls.objects.filter(build=instance):
+            if image.file:
+                files_to_delete.append(image.file.name)
+
+        # Schedule file deletion after transaction is committed
+        if files_to_delete:
+            transaction.on_commit(lambda: delete_firmware_files.delay(files_to_delete))
```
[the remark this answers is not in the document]
youhaveme9: @nemesifier 
This was suggested by Aryaman to use partials

PR #328, opened 2025-07-07
Retry firmware v2

Tosinibikunle:
This PR adds a new API endpoint that allows retrying failed UpgradeOperation instances. The endpoint is intended to improve the operational robustness by giving users the ability to trigger a retry via the API instead of relying solely on automated retries or manual DB intervention. it also adds an update to models.py, task.py, and admin.py.


Tosinibikunle:
@nemesifier  thanks for the advise. I will work on that.

PR #303, opened 2025-04-16
Extended controller apps tests

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [X] I have written new test cases for new code and/or updated existing tests for changes to existing code.
- [ ] I have updated the documentation.

## Reference to Existing Issue

Closes #298 

## Description of Changes

Added sample_config and sample_connection in tests.
Adjust the tests to pass (OK on my machine, still WIP for CI...)

PR #362, opened 2025-12-01
[tests] Extended controller app tests part one

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.
- [X] I have written new test cases for new code and/or updated existing tests for changes to existing code.

This is a part of #303 which was too ambitious. Making a sample_config extend app is trickier than expected, and will be done later.

The goal here is to add a sample_connection extended app, to make sure the case where connection is an extended app is tested.
This happens when someone extends the connection app (to change the connection logic or add complex device command code for instance) and wants to work with it while also using the firmware_upgrader module, which depends on it.

## Description of Changes

Added sample_connection extended app.
Based on #303 changes.
Aiming to get all tests to pass with this as a first step.

To achieve this, we need to refactor some mocks in tests to use a contextmanager for the `*TestCase._mock_connect` attribute instead of a decorator (parsed at parent-class definition time)


nemesifier (commented):
Thanks @asmodehn for your patience in working on this, I scanned the PR and noticed a few small things.

Is the change to `tests/media/floorplan.jpg` intentional? I suggest reverting that.

See also my comments below.

PS: can we close https://github.com/openwisp/openwisp-firmware-upgrader/pull/303 then?

tests/openwisp2/urls.py
```diff
@@ -1,18 +1,64 @@
...
+
 from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.contrib.staticfiles.urls import staticfiles_urlpatterns
 from django.urls import include, path, reverse_lazy
 from django.views.generic import RedirectView
 
+# from openwisp_controller.config.api.urls import get_api_urls as get_config_api_urls
+# from openwisp_controller.config.utils import get_controller_urls
```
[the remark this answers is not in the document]
asmodehn: Actually, it was practical to have here, but I have this information in the other PR, so I removed it from here.

tests/openwisp2/sample_connection/tests.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEqual(n.actor, self.d.deviceconnection_set.first())
+        self.assertEqual(n.target, self.d)
+        self.assertIn(exp_message.format(n=n, target_link=exp_target_link), n.message)
+        self.assertEqual(n.email_subject, exp_email_subject.format(n=n))
+
+
+class TestNotificationTransaction(BaseTestNotificationTransaction):
+    app_label = "sample_connection"
+
+    def _generic_notification_test(
```
nemesifier: This looks identical to the test above, are you sure it's really needed? We use the TransactionTestCase when DB transactions need to be enabled for the feature to work, I think you can remove this test because it won't be doing anything more than the other test method above.

asmodehn: This is probably cleaner : https://github.com/openwisp/openwisp-controller/pull/1192/files
and should simplify things here... Let me know hat you think

tests/openwisp2/sample_connection/tests.py
```diff
@@ -0,0 +1,92 @@
...
+        path = reverse("connection_api:deviceconnection_list", args=(d1.pk,))
+        data = {
+            "credentials": self._get_credentials().pk,
+            "update_strategy": conn_settings.UPDATE_STRATEGIES[0][0],
+            "enabled": True,
+            "failure_reason": "",
+        }
+        with self.assertNumQueries(13):
+            response = self.client.post(path, data, content_type="application/json")
+        self.assertEqual(response.status_code, 201)
```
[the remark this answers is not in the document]
asmodehn: The parent class does not have that field. it is therefore nor required nor useful to have here.

tests/openwisp2/urls.py
```diff
@@ -1,18 +1,46 @@
...
+            include(
+                (
+                    get_connection_api_urls(connection_api_views),
+                    "connection_api",
+                ),
+                namespace="connection_api",
+            ),
+        ),
+    ]
+
```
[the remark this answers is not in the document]
asmodehn: This is the way we test for SAMPLE_APP in other modules, so lets keep it like that and lets not complicate this more than required.

nemesifier:
@coderabbitai full review

tests/openwisp2/sample_connection/api/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+class CredentialListCreateView(BaseCredentialListCreateView):
+    pass
+
+
+class CredentialDetailView(BaseCredentialDetailView):
+    pass
+
+
+class DeviceConnenctionListCreateView(BaseDeviceConnenctionListCreateView):
```
nemesifier: Typo here and on the import alias (line 17) and the as_view calls (lines 49-50). Connenction -> Connection (double n). Also deviceconnection_details_view on line 50 should be detail_view.

tests/openwisp2/sample_connection/pytest.py
```diff
@@ -0,0 +1,10 @@
+from openwisp_controller.connection.tests.pytest import (
```
nemesifier, line 3: It looks like this file is not wired into any test runner.

nemesifier (changes requested):
I scanned the PR and noticed a few things.

The description explains what the code does but not why. What is the real-world use case that requires extending the connection app? Please add that context so it can go into the commit message once merged.

`openwisp_firmware_upgrader/tests/test_admin.py:467` -- The `*args` parameter in `test_add_credentials_with_cancelled_upgrade_operation` is leftover from the old `@mock.patch` decorator pattern. The method now uses context managers via `self._mock_connect` / `self._mock_upgrade`, so `*args` receives nothing. Remove it.

`tests/openwisp2/settings.py` -- Removing the TEMPLATES DIRS entry is unrelated to the connection swap. Was this causing a problem? If not, please revert it to keep the PR focused.

On the mock refactoring -- converting `test_add_credentials_with_cancelled_upgrade_operation` to use `self._mock_connect` / `self._mock_upgrade` is correct. But `TestAdminTransaction` still uses `@mock.patch(_mock_connect, ...)` decorators, which capture the value at parent-class definition time. Subclassing `_mock_connect` will not affect them. This is not a regression but should be acknowledged in the description so reviewers know the scope.

tests/openwisp2/sample_connection/api/views.py
asmodehn: >Also deviceconnection_details_view on line 50 should be detail_view.

I m duplicating this line : https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/connection/api/views.py#L153 so we ll need to fix the typo in multiple places...

asmodehn: Actually the other one is there too : https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/connection/api/views.py#L133

asmodehn:
>On the mock refactoring -- converting test_add_credentials_with_cancelled_upgrade_operation to use self._mock_connect / self._mock_upgrade is correct. But TestAdminTransaction still uses @mock.patch(_mock_connect, ...) decorators, which capture the value at parent-class definition time. Subclassing _mock_connect will not affect them. This is not a regression but should be acknowledged in the description so reviewers know the scope.

Not sure what you mean here, all instances of `_mock_connect_` I could find are in context managers.
Anyway I added a line in the description to be more explicit about this.

tests/openwisp2/sample_connection/pytest.py
asmodehn: Ah good catch. I ll need to review this and add pytest in the test runner

tests/openwisp2/sample_connection/api/views.py
nemesifier: lol! I'll see what I can do

tests/openwisp2/urls.py
```diff
@@ -1,15 +1,42 @@
...
 redirect_view = RedirectView.as_view(url=reverse_lazy("admin:index"))
 
-urlpatterns = [
+urlpatterns = []
+
+if os.environ.get("SAMPLE_APP", False):
+    urlpatterns += [
+        path(
+            "",
+            include(("openwisp_controller.config.urls", "config"), namespace="config"),
```
nemesifier: isn't this redundant with what we have below?

tests/openwisp2/urls.py
```diff
@@ -1,15 +1,42 @@
...
+                ),
+                namespace="connection_api",
+            ),
+        ),
+    ]
+
+urlpatterns += [
     path("admin/", admin.site.urls),
     path("", redirect_view, name="index"),
     path("", include("openwisp_controller.urls")),
```
nemesifier, line 38: Here, doesn't this already include the controller URLs? See https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/urls.py

tests/openwisp2/urls.py
```diff
@@ -1,15 +1,42 @@
...
+    urlpatterns += [
+        path(
+            "",
+            include(("openwisp_controller.config.urls", "config"), namespace="config"),
+        ),
+        path(
+            "api/v1/",
+            include(
+                (
+                    get_connection_api_urls(connection_api_views),
```
nemesifier, line 27: I understand you're doing this to supply your own views, which seems correct, but wouldn't this be overridden by `path("", include("openwisp_controller.urls")),` below?

nemesifier (commented):
Looks good, I have a few doubts touching `urls.py`, I am not 100% sure but worth double checking (I aslso double checking on my side).

nemesifier:
GPT-5.5 assisted review note, focused only on the URL doubts.

The URL concerns look valid:

- `tests/openwisp2/urls.py:23-26` is redundant. `path("", include("openwisp_controller.urls"))` already includes `openwisp_controller.config.urls` under the `config` namespace.
- `path("", include("openwisp_controller.urls"))` does include the controller URL set, but not the swapped `sample_connection` API when `SAMPLE_APP=1`. In `settings.py`, `openwisp_controller.connection` is removed from `INSTALLED_APPS`, so `openwisp_controller.urls` skips its `connection_api` entry.
- The custom `connection_api` include is therefore needed, and it is not overridden by the later `openwisp_controller.urls` include. There is no second `connection_api` namespace in that mode.
- The current mount point looks wrong: `get_connection_api_urls()` already returns paths beginning with `api/v1/controller/...`, so wrapping it in `path("api/v1/", ...)` produces `api/v1/api/v1/controller/...`.

Suggested fix: remove the explicit `openwisp_controller.config.urls` include, keep the custom connection API include, but mount it at `""` rather than `"api/v1/"`.

nemesifier (approved):
@asmodehn I pushed some minor fixes, please double check on your end when you can.

nemesifier:
@coderabbitai resolve

asmodehn:
@nemesifier LGTM :+1: 
In a further effort, we should probably look for a (low maintenance) way to keep the same code in the sample apps here and in controller tests for consistency...

nemesifier:
> @nemesifier LGTM 👍 In a further effort, we should probably look for a (low maintenance) way to keep the same code in the sample apps here and in controller tests for consistency...

You mean shared logic for these sample apps? If it's exactly identical we can do it easily.

PR #365, opened 2025-12-30
[fix] Add data migration to update firmware image for changed image identifiers

### LINKED ISSUE #358: [bug] Crash due to to firmware image identifiers changed
The commit https://github.com/openwisp/openwisp-firmware-upgrader/commit/1749ca0c623f39420b5e6b6fe6785dbf4c39bd81 introduced a few changes that can break existing instance, eg: "x86-generic-combined-squashfs.img.gz" was changed in OpenWrt, so I updated it there, but forgot that existing instance will crash if the value in the DB is not updated.

I think the solution is to write a data migration which would automatically update any firmware image using one of the image identifiers that has been changed.

youhaveme9:
## Checklist

Closes #358

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -2,4 +2,4 @@ openwisp-utils[qa,selenium] @ https://github.com/openwisp/openwisp-utils/archive
 django-redis~=6.0.0
 mock-ssh-server~=0.9.1
 responses~=0.25.8
-psycopg2-binary~=2.9.10
+psycopg2-binary~=2.9.11


tests/openwisp2/sample_firmware_upgrader/migrations/0005_update_image_type_identifiers.py
```diff
@@ -0,0 +1,29 @@
+from django.db import migrations
```
nemesifier: we usually don't do data migrations in sample apps, we just erase the DB and start from scratch because it's only for testing.

nemesifier (commented):
Looks good, see my comments below aimed at reducing maintenance overhead.

PR #384, opened 2026-02-27
[fix] Prevent upgrade operations on deactivated devices #382

### LINKED ISSUE #382: [bug] Upgrade operations should not include deactivated devices
**Describe the bug**
Currently, the firmware upgrader allows creation of upgrade operations for devices that are deactivated. Validation should fail when attempting to create upgrade operations for such devices. Additionally, mass upgrade operations must exclude deactivated devices in the query that generates the upgrade operations.

**Steps To Reproduce**

1. Deactivate a device via the admin or API.
2. Attempt to create an upgrade operation for the deactivated device (manually or via mass upgrade).
3. Observe that upgrade operations are incorrectly allowed or included for deactivated devices.

**Expected behavior**
Upgrade operations must not be allowed for deactivated devices.
- Validation should fail if an upgrade operation is requested for a deactivated device.
- Mass upgrade queries must exclude all deactivated devices.
- Comprehensive model tests, admin tests (test_admin), and API tests (test_api) must be created to verify this behavior.

SAMurai-16:
Added comprehensive validation to prevent firmware upgrade operations on deactivated devices across all application layers:

- Added model validation in AbstractDeviceFirmware.clean() and AbstractUpgradeOperation.clean() to reject operations on deactivated devices
- Enhanced API validation in DeviceFirmwareSerializer to validate device state
- Updated batch operation filtering in _find_related_device_firmwares() and _find_firmwareless_devices() to exclude deactivated devices
- Added comprehensive tests for model, API, and admin validation

This ensures defense-in-depth protection where deactivated devices cannot be targeted for firmware upgrades through any interface (UI, API, or batch operations), preventing potential security issues.

Related to https://github.com/openwisp/openwisp-controller/issues/1338
Fixes #382

## Checklist

Closes #382 .


openwisp_firmware_upgrader/base/models.py
```diff
@@ -206,6 +206,7 @@ def _find_related_device_firmwares(
             .select_related(*related)
             .filter(image__build__category_id=self.category_id)
             .exclude(image__build=self, installed=True)
+            .exclude(device___is_deactivated=True)
```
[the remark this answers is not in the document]
nemesifier: Three underscore can't be possibly working, please double check. The fact that tests are passing is a BIG RED FLAG.

nemesifier (changes requested):
There's a few coderabbitai comments which were ignored, please address  them.

### openwisp_firmware_upgrader/tests/test_api.py (1)

`2124-2137`: **Potential issue | Major**

**This “API” test still exercises only model validation, not the API path.**

On Line 2124-Line 2134, the test only calls `full_clean()` and never hits an endpoint, so it does not validate API response semantics (status/body) for deactivated devices.

#### Suggested fix (exercise real endpoint instead of model-only validation)

```diff
 class TestDeactivatedDeviceAPI(TestAPIUpgraderMixin, TestCase):
@@
     def test_device_firmware_basic_deactivated_validation(self):
-        """Test that DeviceFirmware creation is rejected for deactivated devices"""
-        device_fw = self._create_device_firmware()
-        device = device_fw.device
-        device.deactivate()
-
-        # Test that we cannot create new device firmware for deactivated device
-        with self.assertRaises(ValidationError) as cm:
-            new_device_fw = DeviceFirmware(device=device, image=device_fw.image)
-            new_device_fw.full_clean()
-
-        self.assertIn(
-            "Cannot create firmware object for deactivated device", str(cm.exception)
-        )
+        """Test API rejects DeviceFirmware creation for deactivated devices."""
+        env = self._create_upgrade_env(device_firmware=False)
+        device = env["d1"]
+        image = env["image1a"]
+        device.deactivate()
+
+        url = reverse("upgrader:api_devicefirmware_detail", args=[device.pk])
+        response = self.client.put(
+            url,
+            {"image": image.pk},
+            content_type="application/json",
+        )
+        self.assertEqual(response.status_code, 400)
+        self.assertIn(
+            "Cannot create firmware object for deactivated device",
+            str(response.data),
+        )

```

---

Would you like me to help you apply these changes to your local file or run a specific validation script?

There's also a bunch of copiliot comments you didn't address. Please respond to each: if the suggestion is valid, change the code accordingly, if not, explain why the suggestion is invalid.

### Manual testing

Please confirm you've been manually testing this to ensure all the problems described in the issue are resolved. Screen recordings that demonstrate this help.
We're getting a lot of untested patches which are overloading us and I have no other way than this to avoid us wasting time on untested coded. The fact that there's a HUGE RED FLAG in your code (a queryset ORM which uses three underscores and tests are passing anyway) doesn't convince me in spending further time on this.

SAMurai-16:
@nemesifier The field is defined as "_is_deactivated" in openwisp-controller device model and "device__ " is being used for django ORM traversal 
So i believe "device___is_deactivated" is the correct Django ORM lookup.

See [this implementation](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/base/device.py#L109-L111) for reference.

nemesifier:
@coderabbitai review

SAMurai-16:
@nemesifier if any changes are required then do tell

openwisp_firmware_upgrader/api/views.py
```diff
@@ -66,6 +67,29 @@ def get_queryset(self):
         return qs
 
 
+class RelatedDeviceAPIMixin(ProtectedAPIMixin):
```
pandafy, line 65: Why is this re-implemented here?

openwisp_firmware_upgrader/base/models.py
```diff
@@ -395,8 +396,12 @@ class Meta:
         abstract = True
 
     def clean(self):
-        if not hasattr(self, "image") or not hasattr(self, "device"):
+        if not self.device_id or not self.image_id:
             return
```
pandafy, line 404: This change looks unneccessary.

pandafy (changes requested):
**Note:** AI assisted review of the current PR. 

## Does it solve the issue

Yes, on the functional level. The two things the issue asks for are both covered:

- Creating or modifying a firmware/upgrade object for a deactivated device is now
  rejected at several layers (model `clean()`, the API serializer, the API
  permission class, and the admin).
- The mass upgrade queries now skip deactivated devices. Both
  `_find_related_device_firmwares()` and `_find_firmwareless_devices()` exclude
  them, and since `dry_run()` and the actual upgrade run through those same two
  methods, the preview and the execution stay consistent.

I verified the new model, admin and API tests pass on their own, and the batch
exclusion works for both the firmwareless and the non-firmwareless path.

The problem is the implementation of the API permission rework. It is the last
commit (`a8314c1`) and it breaks the test suite. Details below.

## Blocking

### 1. The test suite is red, all CI jobs are failing

All 12 matrix jobs on the PR are failing, and they fail locally too. Running the
full `test_api` module gives 10 failures, every one of them an
`assertNumQueries` mismatch in `TestDeviceFirmwareImageViews`. These same tests
pass on `master`, so this is a regression introduced by the new
`RelatedDeviceAPIMixin`.

The extra queries come from the new permission machinery. Some examples of the
jump:

- `test_device_firmware_detail_create`: 26 -> 33 queries
- `test_device_firmware_detail_create_shared_image`: 25 -> 32
- `test_device_firmware_detail_update`: 27 -> 32
- `test_device_firmware_detail_get`: 9 -> 13 and 8 -> 12
- `test_device_firmware_detail_multitenancy`: 6 -> 10

Worth noting that even the plain GET tests gained queries. That is because those
tests hit the browsable API (`format=api`), and DRF renders the write form by
running a cloned permission check, which now triggers the parent device lookup.
So the new permission leaks into read requests through the browsable renderer.

The create path is the worst (+7) because the parent device gets looked up more
than once for a single request: once in `has_permission` on the original PUT,
again in `assert_parent_exists()`, and again in the cloned POST permission check
inside `get_object_or_none()`. None of those results are reused.

## Should be fixed

### 2. The new permission stacks on top of the existing one instead of replacing it

`ProtectedAPIMixin` already includes `DjangoModelPermissions` in its
`permission_classes`. `RelatedDeviceModelPermission` is itself a subclass of
`DjangoModelPermissions`, and `get_permissions()` appends it on top for write
methods. So a write request now runs the model permission check twice. That is
part of why the query count went up, and it is just redundant work. Replacing
the model permission in the list with `RelatedDeviceModelPermission` (the way
`RelatedDeviceProtectedAPIMixin` does upstream) avoids running the same check
twice.

### 3. Grabbing the permission class by list index is fragile

```python
related_device_permission_class = RelatedDeviceProtectedAPIMixin.permission_classes[-1]
```

This depends on `RelatedDeviceModelPermission` staying the last entry of an
upstream list. If openwisp-controller ever reorders that list the wrong class
gets pulled in, with no error to tell you. The class is importable, so just
import it directly:

```python
from openwisp_controller.mixins import RelatedDeviceModelPermission
```

### 4. `assert_parent_exists` uses a bare assert for control flow

```python
try:
    assert parent_queryset.exists()
except (AssertionError, ValidationError):
    raise NotFound(detail="device not found")
```

Running Python with `-O` strips assert statements, so under that flag
`parent_queryset.exists()` is never evaluated and the `NotFound` is never raised.
The behaviour of a PUT-as-create against a missing device would silently change.
Use a normal conditional:

```python
if not parent_queryset.exists():
    raise NotFound(detail="device not found")
```

You still want the `ValidationError` guard for the invalid-UUID case, so wrap the
`.exists()` call in try/except for that, but do not rely on `assert`.

## Minor

### 5. The serializer deactivation check is now dead for the API

`DeviceFirmwareSerializer.validate()` raises "Cannot create or modify firmware
object for deactivated device", but with the new permission class a write request
on a deactivated device returns 403 before the serializer ever runs. The model
`clean()` covers the defense-in-depth case anyway. The serializer block is not
harmful, but it is unreachable through the API as the code stands. We need to a
dedicated test for the serializer.


### 6. The firmwareless batch test only checks exclusion, not inclusion

In `test_batch_upgrade_excludes_deactivated_devices`, the firmwareless=True part
only asserts the deactivated device is absent. It never asserts that an active
firmwareless device is present. The test would still pass if the firmwareless
query were broken to return nothing. The firmwareless=False part does check both
sides, so only the firmwareless path is under-covered. Add an active firmwareless
device and assert it shows up in the operations.

### 7. Unrelated change in the diff

`.gitignore` adds `.venv/`. Harmless, but it is unrelated to the fix and just
adds noise to the PR.

nemesifier:
I went ahead and prepared a maintainer-side fix for the remaining issues here.

The important gaps were:

- Queued upgrade operations were still able to run if the device was deactivated after the operation was created but before the worker started. The worker path now aborts before opening a device connection and records the reason in the operation log.
- The API returned a generic 403 for deactivated-device firmware writes. It now returns a clear error message: `Firmware upgrades are not allowed for deactivated devices.`
- The deactivated-device API permission flow was doing an avoidable parent-device lookup. The implementation now keeps the existing query-count expectations stable.
- The batch-upgrade preview API now has coverage to ensure deactivated devices are excluded there too, not only during actual batch creation.

This keeps the fix focused on preventing firmware upgrades from running against deactivated devices at creation time, API write time, batch selection time, and worker execution time.

nemesifier:
@coderabbitai full review

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai review

nemesifier (approved):
Thanks @SAMurai-16 @pandafy 👏

PR #397, opened 2026-03-10
[tests] Replace global mock decorator with local contextmanager in test_admin.TestAdminTransaction

asmodehn:
## Checklist

- [X] I have read the [OpenWISP Contributing Guidelines](http://openwisp.io/docs/developer/contributing.html).
- [X] I have manually tested the changes proposed in this pull request.

## Description of Changes

Refactoring one test file to allow for simple inheritance.
Part of #362  for simpler check

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_admin.py
+++ b/openwisp_firmware_upgrader/tests/test_admin.py
@@ -652,898 +652,939 @@ class TestAdmin(BaseTestAdmin, TestCase):
             )
 
 
-_mock_upgrade = "openwisp_firmware_upgrader.upgraders.openwrt.OpenWrt.upgrade"
-_mock_connect = "openwisp_controller.connection.models.DeviceConnection.connect"
-
-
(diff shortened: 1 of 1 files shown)


nemesifier (commented):
@asmodehn the change looks ok to merge, I don't see any problem with it, but I do not understand the need for this change and I think the pull request description fails to explain this. Why is this change needed, what does the current code prevent you to do?

asmodehn:
@nemesifier The issue is that the `@mock.patch`, when used as a decorator cannot be overloaded from extended tests. It is applied at class parsing time, when the field value is the one from original test class in openwisp module.
The context manager usage however uses a value that can be overloaded via a field in an extended test class.

It is the same as the long going PR I did before : #303 and #362. I just figured out I'd do the change one file / one test class as a time as the change makes a huge diff for little reason...

nemesifier (approved):
Thanks for the explanation @asmodehn.

PR #401, opened 2026-03-24
[fix] Prevent 500 and warn about credentialless devices on mass upgrade #302

### LINKED ISSUE #302: [bug] Mass upgrade results in 500 internal server error if there's any device with no credentials assigned
**Describe the bug**

Trying to initiate a mass upgrade if there's a device which does not have a device credentials assigned (`DeviceConnection` instance) raises an uncaught exception and results in 500 internal server error.

Similar to #250, but triggered from mass upgrades.

**Steps To Reproduce**

1. Create a category, build and a firmware image
2. Create a device with related credentials and a firmware image object (firmware image created at step 1)
3. Delete the device credentials (go to device change page, credentials tab, flag the delete checkbox, save)
4. Go to the firmware build section, select the build created at step 1, launch mass upgrade operation 

Result: uncaught exception.
Stack trace:

Stack trace:
```
Traceback (most recent call last):
  File "/lib/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/lib/python3.10/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/lib/python3.10/site-packages/django/contrib/admin/options.py", line 688, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/lib/python3.10/site-packages/django/utils/decorators.py", line 134, in _wrapper_view
    response = view_func(request, *args, **kwargs)
  File "/lib/python3.10/site-packages/django/views/decorators/cache.py", line 62, in _wrapper_view_func
    response = view_func(request, *args, **kwargs)
  File "/lib/python3.10/site-packages/django/contrib/admin/sites.py", line 242, in inner
    return view(request, *args, **kwargs)
  File "/lib/python3.10/site-packages/django/utils/decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "/lib/python3.10/site-packages/django/utils/decorators.py", line 134, in _wrapper_view
    response = view_func(request, *args, **kwargs)
  File "/lib/python3.10/site-packages/django/contrib/admin/options.py", line 1959, in changelist_view
    response = self.response_action(
  File "/lib/python3.10/site-packages/django/contrib/admin/options.py", line 1593, in response_action
    response = func(self, request, queryset)
  File "/openwisp-firmware-upgrader/openwisp_firmware_upgrader/admin.py", line 187, in upgrade_selected
    upgrader_schema = BatchUpgradeOperation(build=build)._get_upgrader_schema(
  File "/openwisp-firmware-upgrader/openwisp_firmware_upgrader/base/models.py", line 569, in _get_upgrader_schema
    upgrader_class = self._get_upgrader_class(
  File "/openwisp-firmware-upgrader/openwisp_firmware_upgrader/base/models.py", line 561, in _get_upgrader_class
    return get_upgrader_class_for_device(related_device_fw.first().device)
  File "/openwisp-firmware-upgrader/openwisp_firmware_upgrader/utils.py", line 32, in get_upgrader_class_for_device
    raise device.deviceconnection_set.model.DoesNotExist

Exception Type: DoesNotExist at /admin/firmware_upgrader/build/
Exception Value: 
```

**Expected behavior**

I wonder if we can catch the exception and list any device which doesn't have any credentials specified at the end of the intermediate page, saying something like the system can't fix this issue and the user shall assign credentials to those devices or something similar.

Adityashandilya555:
## Problem

When triggering a mass firmware upgrade, if any device (or all devices) lacked a `DeviceConnection` with SSH credentials, `_get_upgrader_class()` would call `get_upgrader_class_for_device()` on an uncredentialed device, raising an unhandled `DoesNotExist` and returning a 500 error.

Additionally, there was no UI feedback to inform administrators which devices would be skipped due to missing credentials.

### `base/models.py` — `_get_upgrader_class()`
- Wraps the `upgrade_operations` path in `try/except ObjectDoesNotExist`
- Uses an ORM subquery to find the **first device with a valid SSH credential** from both `related_device_fw` and `firmwareless_devices`, instead of blindly using the first device regardless of credentials

### `admin.py` — `upgrade_selected()`
- After `dry_run()`, identifies devices without valid SSH credentials from both querysets using ORM exclusion
- Builds a list of up to `MAX_CREDENTIALLESS_DISPLAY` (5) devices with their admin change URLs
- Calculates overflow count (`credentialless_more`) for the "and N more" message
- Passes `credentialless_devices`, `credentialless_count`, and `credentialless_more` to the template context

### `templates/admin/upgrade_selected_confirmation.html`
- New `errornote` warning block shown when `credentialless_count > 0`
- Lists each credentialless device as a clickable admin link
- Shows "... and N more device(s) without credentials (not shown)." when the list is truncated

### `tests/test_admin.py`
- `test_upgrade_intermediate_page_credentialless_related` — credentialless device in related DeviceFirmware
- `test_upgrade_intermediate_page_credentialless_firmwareless` — credentialless firmwareless device
- `test_upgrade_intermediate_page_all_credentialless` — all devices lack credentials, no 500
- `test_upgrade_intermediate_page_credentialless_more_than_max` — overflow beyond display limit triggers "and N more"

## Fixes

Fixes #302

### DIFF
--- a/openwisp_firmware_upgrader/admin.py
+++ b/openwisp_firmware_upgrader/admin.py
@@ -40,6 +40,7 @@ from .utils import get_upgrader_schema_for_device
 from .widgets import FirmwareSchemaWidget, MassUpgradeSelect2Widget
 
 logger = logging.getLogger(__name__)
+MAX_CREDENTIALLESS_DISPLAY = 5
 BatchUpgradeOperation = load_model("BatchUpgradeOperation")
 UpgradeOperation = load_model("UpgradeOperation")
 DeviceFirmware = load_model("DeviceFirmware")
@@ -261,6 +262,45 @@ class BuildAdmin(BaseAdmin):
         )
         related_device_fw = result["device_firmwares"]
         firmwareless_devices = result["devices"]
+        # Identify devices that lack valid SSH credentials so admins are informed.
+        valid_device_ids = DeviceConnection.objects.filter(
+            update_strategy__icontains="ssh",
+            enabled=True,
+        ).values("device_id")
+        credentialless_fw = related_device_fw.exclude(device_id__in=valid_device_ids)
+        credentialless_firmwareless = firmwareless_devices.exclude(
+            pk__in=valid_device_ids
+        )
+        total_credentialless = (
+            credentialless_fw.count() + credentialless_firmwareless.count()
+        )
+        device_app_label = Device._meta.app_label
+        credentialless_devices = []
+        for df in credentialless_fw.select_related("device")[
+            :MAX_CREDENTIALLESS_DISPLAY
+        ]:
+            credentialless_devices.append(
+                {
+                    "name": df.device.name,
+                    "url": reverse(
+                        f"admin:{device_app_label}_device_change",
+                        args=[df.device_id],
+                    ),
+                }
+            )
+        remaining = MAX_CREDENTIALLESS_DISPLAY - len(credentialless_devices)
+        if remaining > 0:
+            for device in credentialless_firmwareless[:remaining]:
+                credentialless_devices.append(
+                    {
+                        "name": device.name,
+                        "url": reverse(
+                            f"admin:{device_app_label}_device_change",
+                            args=[device.pk],
+                        ),
+                    }
+                )
+        credentialless_more = max(0, total_credentialless - MAX_CREDENTIALLESS_DISPLAY)
         title = _("Confirm mass upgrade operation")
         context = self.admin_site.each_context(request)
         upgrader_schema = BatchUpgradeOperation(build=build)._get_upgrader_schema(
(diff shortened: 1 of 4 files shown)


SAMurai-16:
Hi @Adityashandilya555 i was going through your approach and i had a question 
At what part is the `credentialless_count` being calculated in your approach?

nemesifier (changes requested):
Thanks @Adityashandilya555, this is a real improvement. Preventing the 500 from the unhandled `DoesNotExist` in `_get_upgrader_class()` and showing admins which devices will be skipped is good UX for #302.

A few things to address:

1. **The SSH filter is a fragile heuristic.** `update_strategy__icontains="ssh"` relies on the connector's dotted path containing the substring "ssh". That works for the built-ins, but a custom SSH-based connector whose path lacks "ssh", or a non-SSH path that happens to contain it, would be misclassified. Consider checking the connector's actual capability rather than a substring match, or at least add a comment documenting the assumption.
2. **Guard the root cause too (TOCTOU).** Pre-filtering at dry-run/confirm time is good, but a device can lose its connection between confirm and execution. To be robust, `_get_upgrader_class()` (or the execution path) should also handle the missing-connection case gracefully instead of 500ing. The pre-filter improves UX; the guard prevents the crash for good.
3. **Duplication.** The credentialless filtering logic appears in both `admin.py` and `base/models.py`. Please factor it into a single helper (e.g. on the model) so the admin and the batch logic cannot drift apart.

Good direction. Address the source-level guard (point 2) so the 500 truly cannot happen, tidy the duplication, and we are good.

PR #403, opened 2026-03-24
[fix] Remove invalid escape sequence warnings in OpenWrt strings

yassinekolsi:
Summary
This PR removes Python 3.12+ invalid escape sequence SyntaxWarnings in OpenWrt upgrader/test strings.

Changes
- Converted OpenWrt false-positive regex patterns to raw strings to avoid invalid escape warnings.
- Updated false-positive regex patterns to match both escaped and unescaped path variants (e.g. \/tmp and /tmp), preserving intended suppression behavior.
- Simplified OpenWrt test fixture strings by removing unnecessary escaped forward slashes.

Behavior impact
No intended functional behavior change beyond keeping false-positive matching robust across escaped/unescaped log formats.

Validation
- Before (warnings reproduced):
  - python3 -W default -m py_compile openwisp_firmware_upgrader/upgraders/openwrt.py openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
- After (clean):
  - python3 -W error::SyntaxWarning -m py_compile openwisp_firmware_upgrader/upgraders/openwrt.py openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py

Checklist
- [x] Read contributing guidelines.
- [x] Manually validated changes.
- [x] Updated test fixtures as needed.
- [ ] Documentation update not needed for this housekeeping fix.

### DIFF
--- a/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
+++ b/openwisp_firmware_upgrader/tests/test_openwrt_upgrader.py
@@ -192,10 +192,10 @@ def mocked_exec_upgrade_success_false_positives(
         filename = command.split()[-1].split("/")[-1]
         raise CommandFailedException(
             "Command failed: ubus call system sysupgrade "
-            '{ "prefix": "\/tmp\/root", '
-            f'"path": "\/tmp\/{filename}", '
-            '"backup": "\/tmp\/sysupgrade.tgz", '
-            '"command": "\/lib\/upgrade\/do_stage2", '
+            '{ "prefix": "/tmp/root", '
+            f'"path": "/tmp/{filename}", '
+            '"backup": "/tmp/sysupgrade.tgz", '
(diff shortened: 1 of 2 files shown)


nemesifier (commented):
Thanks @yassinekolsi, this is a correct cleanup. Converting the `_false_positives` regex patterns to raw strings removes the Python 3.12 `SyntaxWarning`s, and using `\?/tmp` so the patterns match both escaped and plain slashes is a nice touch that keeps the suppression working regardless of how the device emits the path. Simplifying the test fixture to plain slashes is consistent with that.

One small suggestion:

- Since the regex now intentionally handles both escaped and unescaped slashes, consider keeping (or adding) one test case for each form so the `\?` optionality is actually exercised. Right now the fixture only covers the plain variant.

Otherwise good to merge.

PR #447, opened 2026-06-13
[fix] Guard admin reverse and template url behind FIRMWARE_UPGRADER_API #444

### LINKED ISSUE #444: [bug] Admin reverses upgrader: URLs unconditionally, ignoring FIRMWARE_UPGRADER_API
**Describe the bug**

The device admin change page and the upgrade-operation admin change page reverse `upgrader:api_upgradeoperation_cancel` unconditionally, but the URL that name resolves to is only registered when the API is enabled.

`openwisp_firmware_upgrader/urls.py` guards the API urls behind the setting:

```python
if app_settings.FIRMWARE_UPGRADER_API:
    urlpatterns += [
        path("api/v1/", include("openwisp_firmware_upgrader.api.urls")),
    ]
```

`api/urls.py` is what carries `app_name = "upgrader"`, so when `OPENWISP_FIRMWARE_UPGRADER_API=False` the `upgrader` namespace is never registered. However two places in the admin reverse it regardless of the setting:

- `openwisp_firmware_upgrader/admin.py:438` (`UpgradeOperationAdmin.change_view`)
- `openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html:8` (rendered by `DeviceFirmwareInline` on the device change page)

Both were added in 1.3 with the real-time upgrade progress feature (#224). With the API disabled, the reverse raises `NoReverseMatch` and the affected admin pages return HTTP 500.

The default is `FIRMWARE_UPGRADER_API=True`, so the common case is unaffected. The bug only surfaces when the API is explicitly disabled.

**Steps To Reproduce**

1. Set `OPENWISP_FIRMWARE_UPGRADER_API = False` in the project settings.
2. Open the change page of any device that shows the firmware inline (or open an upgrade operation change page).
3. The page returns HTTP 500 with `NoReverseMatch` for `upgrader:api_upgradeoperation_cancel`.

**Expected behavior**

The admin should not depend on URLs that are conditionally registered. When `FIRMWARE_UPGRADER_API` is disabled, the device and upgrade-operation admin pages should render without error (the cancel-upgrade UI can be omitted or the reverse guarded behind the same setting).

**System Informatioon:**

- OS: Ubuntu 24.04 LTS
- Python Version: Python 3.14
- Django Version: Django 5.2
- openwisp-firmware-upgrader: 1.3 dev (master)
- Browser and Browser Version (if applicable): N/A

BHARATH0153:
## Checklist

Closes #444 

## Changes

Guard the `reverse("upgrader:api_upgradeoperation_cancel")` call in the admin and the `{% url %}` tag in the template behind `FIRMWARE_UPGRADER_API`. When the API is disabled, the URL doesn't exist and these cause `NoReverseMatch` — now they gracefully return empty string instead of crashing.

### DIFF
--- a/openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html
+++ b/openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html
@@ -5,5 +5,5 @@ var firmwareUpgraderSchema = {{ inline_admin_formset.formset.extra_context | def
 var owFirmwareUpgraderApiHost = {
     host: "{{ request.get_host|escapejs }}"
 };
-var owUpgradeOperationCancelUrl = "{% url 'upgrader:api_upgradeoperation_cancel' '00000000-0000-0000-0000-000000000000' %}";
+var owUpgradeOperationCancelUrl = "{% url 'upgrader:api_upgradeoperation_cancel' '00000000-0000-0000-0000-000000000000' as cancel_url %}{{ cancel_url|default:'' }}";
 </script>
--- a/openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html
+++ b/openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html
@@ -5,5 +5,5 @@ var firmwareUpgraderSchema = {{ inline_admin_formset.formset.extra_context | def
 var owFirmwareUpgraderApiHost = {
     host: "{{ request.get_host|escapejs }}"
 };
-var owUpgradeOperationCancelUrl = "{% url 'upgrader:api_upgradeoperation_cancel' '00000000-0000-0000-0000-000000000000' as cancel_url %}{{ cancel_url|default:'' }}";
+var owUpgradeOperationCancelUrl = "{{ inline_admin_formset.formset.upgrade_operation_cancel_url }}";
 </script>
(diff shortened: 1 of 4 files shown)


BHARATH0153:
@coderabbitai review

BHARATH0153:
@coderabbitai any outstanding issues?

nemesifier (changes requested):
A failing test would have been nice, it should be possible to achieve this by mocking app_settings.

Please clarify why the entire checklist is empty in the PR description, you should flag at least the first two checkboxes. Have you manually tested this?

BHARATH0153:
@nemesifier please review when you are free thanks!

CodingWithSaksham:
@coderabbitai review

openwisp_firmware_upgrader/templates/admin/firmware_upgrader/device_firmware_inline.html
```diff
@@ -5,5 +5,5 @@
 var owFirmwareUpgraderApiHost = {
     host: "{{ request.get_host|escapejs }}"
 };
-var owUpgradeOperationCancelUrl = "{% url 'upgrader:api_upgradeoperation_cancel' '00000000-0000-0000-0000-000000000000' %}";
+var owUpgradeOperationCancelUrl = "{{ inline_admin_formset.formset.upgrade_operation_cancel_url }}";
```
nemesifier, line 8: With the API disabled, this becomes an empty string, but `updateStatusWithProgressBar` still renders an enabled Cancel button. Clicking it reaches `cancelUpgradeOperation`, which POSTs to the empty URL, that is, the current admin change page. The cancellation cannot work and the request is handled as an invalid admin form submission. Please omit or disable the cancel control when this URL is unavailable, and add a regression test for that path.

nemesifier (changes requested):
I cannot approve this while the API-disabled admin page still exposes a cancel action that cannot work.

nemesifier (approved):

nemesifier:
@coderabbitai resolve

Issue #380, opened 2026-02-09
[feature] Scheduled Mass Upgrades


nemesifier:
## Problem statement

Currently, mass firmware upgrades start immediately after the upgrade operation is created. This is problematic in production environments where devices must remain available during business hours. Network operators need a way to plan upgrades during low usage windows, for example overnight or during scheduled maintenance periods, without manual intervention at execution time.

## Describe the solution you'd like

```mermaid
flowchart TD
    A[User creates mass upgrade] --> B{Schedule datetime provided?}

    B -- No --> C[Execute immediately, existing behavior]
    C --> D[Mass upgrade runs via Celery tasks]
    D --> E[Upgrade completes]
    E --> F[Send generic notification]

    B -- Yes --> G[Validate scheduled datetime]
    G -- Invalid --> H[Reject with validation error]
    G -- Valid --> I[Create operation with status SCHEDULED and store UTC datetime]

    I --> J[Periodic scheduler via Celery Beat every minute]
    J --> K{Upgrade due?}

    K -- No --> J
    K -- Yes --> L[Transition status from SCHEDULED to RUNNING]

    L --> M[Runtime validation of permissions firmware and devices]

    M -- All invalid --> N[Cancel operation with FAILED status and error]
    N --> O[Send generic notification for failure]

    M -- Some or all valid --> P[Execute mass upgrade using existing logic]
    P --> Q[Upgrade completes]
    Q --> R[Send generic notification for completion]
```

Introduce support for **scheduled mass upgrades**, allowing a mass upgrade operation to start at a user defined datetime in the future.

Scheduled upgrades must be reliable, persistent across restarts, and clearly communicate timing semantics and execution state to users.

### UI

* On the existing mass upgrade confirmation page, allow users to optionally schedule the upgrade for a future datetime.
* By default, the upgrade executes immediately unless a future datetime is explicitly set.
* The scheduled datetime must:
  * be in the future
  * respect a configurable minimum delay (eg 10 minutes, to avoid accidental scheduling)
  * not exceed a configurable maximum horizon (eg default 6 months)
* Invalid scheduling values must be rejected with a validation error.
* Only upgrades in `scheduled` status are editable. Once execution starts, the operation becomes read only.
* Timezone handling must be explicit:
  * User input should be interpreted in the user browser timezone (or another clearly justified UX choice).
  * Datetimes must be stored in UTC.
  * The server preferred timezone should be clearly indicated in the UI.

### Backend / REST API

[...]

PR #460, opened 2026-06-30
[feature] Scheduled Mass Upgrades

Eeshu-Yadav:
## Checklist

- [x] I have updated the documentation.

Related to #380 (Scheduled Mass Upgrades, Phase 2).

This builds on the persistent-upgrades work, so for now it's based on `issues/417-persistence-schema-fields` and the diff only shows the scheduling changes. I'll retarget it to `gsoc26-persistent-scheduled-upgrades` once persistence merges.

## Description of Changes

This adds the ability to schedule a mass upgrade for a later maintenance window instead of running it immediately, and to edit or cancel it while it waits.

The batch model gets a `scheduled_at` time, a stored `firmwareless` flag and a new `scheduled` status, with a `(status, scheduled_at)` index for the scan. When you pick a future time on the confirmation page, `batch_upgrade()` saves the batch as `scheduled` instead of dispatching it. A Celery Beat task, `execute_scheduled_upgrades`, then picks up batches whose time has come, re-checks that there are still devices to upgrade, and launches them; if nothing is eligible by then the batch is marked failed without touching a device.

To stop two upgrades from fighting over the same devices, creating a batch that overlaps an active one (or a device that is already mid-upgrade) is rejected, with the conflicting operation named in the error.

While a batch is scheduled you can change the time, the group/location filters, and the persistent and firmwareless options, or cancel it. This works both from the admin (Edit Schedule and Cancel buttons on the batch page) and from the REST API (`reschedule/` and `cancel/` endpoints); both go through the same code so they stay in sync. Times are entered in the browser timezone and converted to UTC on submit, the admin shows them back in the server timezone, and the API uses ISO 8601 with an offset.

Notifications go out when a scheduled upgrade starts, when it can't start because no devices are eligible, and when any mass upgrade finishes.

Two settings bound the schedule: `OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MIN_DELAY` (10 minutes) and `OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MAX_HORIZON` (180 days).

Docs are a new `scheduled-mass-upgrades.rst` page plus updates to the REST API, settings and changelog pages. Migrations are additive for both the default and the sample app, so existing rows need no data migration. Screenshots and the demo video are on the `docs` branch, like the persistent-upgrades ones.


nemesifier:
@coderabbitai review

Eeshu-Yadav:
@coderabbitai full review, focus on security, correctness, race conditions, efficiency

Eeshu-Yadav:
@coderabbitai full review, focus on security, correctness, race conditions, efficiency

docs/user/rest-api.rst
```diff
@@ -236,6 +287,18 @@ Example with filters:
...
+
+.. code-block:: json
+
+    {
+        "scheduled_at": "{future_iso8601_datetime}"
+    }
+
+The ``scheduled_at`` value must be a timezone-aware ISO 8601 timestamp in
+the future that satisfies the configured minimum delay and maximum
+scheduling horizon (see :doc:`scheduled-mass-upgrades`).
```
nemesifier: The API contract says an offset is required, but the serializer accepts naive ISO values and silently interprets them in `settings.TIME_ZONE`; `test_reschedule_naive_scheduled_at_interpreted_in_server_timezone` codifies that behavior. A client omitting the offset can therefore schedule the upgrade hours away from its intended instant despite following an endpoint that otherwise appears to use UTC semantics.

Please make the contract consistent. Since issue #380 explicitly requires timezone-aware API values, reject naive datetimes in both create and reschedule requests and update the test accordingly.

Severity: P2

docs/user/scheduled-mass-upgrades.rst
```diff
@@ -0,0 +1,113 @@
...
+Three notifications keep operators informed, all delivered to the
+organization's administrators and superusers:
+
+- a **started** notification when a scheduled batch reaches its time and
+  begins,
+- a **not started** notification when a scheduled batch is skipped because
+  no eligible device remained when its time arrived, and
+- a **completed** notification when a mass upgrade that has started
+  reaches a terminal state, reporting whether it succeeded, failed or was
+  cancelled.
```
nemesifier: The documented completion notification for cancelled batches is not implemented. `notify_on_completion()` accepts only `success` and `failed`, and `test_notifies_noncancelled_completion` explicitly expects zero notifications when an in-progress batch finishes as `cancelled`. This also conflicts with issue #380's requirement to notify when any mass upgrade completes.

Did we discuss this in one of  our meetings? Or what? There's a discrepancy between the plan and the implementation.

Severity: P2

docs/user/scheduled-mass-upgrades.rst
```diff
@@ -0,0 +1,113 @@
...
+Enabling from the admin
+-----------------------
+
+The mass-upgrade confirmation page, reached from a build's *Upgrade*
+action, has a **Scheduled at** field. Leave it empty to upgrade
+immediately, which is the default, or pick a date and time to run the
+batch later.
+
+.. image:: https://raw.githubusercontent.com/openwisp/openwisp-firmware-upgrader/docs/docs/images/1.4/scheduled-mass-upgrade-confirm.png
+    :target: https://raw.githubusercontent.com/openwisp/openwisp-firmware-upgrader/docs/docs/images/1.4/scheduled-mass-upgrade-confirm.png
```
nemesifier, line 50: The documentation assets are not available. All three new raw image URLs return 404 from the `docs` branch, and neither this page nor the PR description links the short usage video required by issue #380. The published page would contain broken images and the media acceptance criteria remain unmet.

Please add the three screenshots to the referenced `docs` branch paths, add the usage-video link, and verify every URL before merging.

Severity: P2

docs/user/settings.rst
```diff
@@ -129,6 +129,37 @@ The Beat task that drives these reminders
...
+============ =======================
+
+Furthest into the future, in seconds, that a mass upgrade may be
+scheduled. A time beyond it is rejected with *The scheduled time cannot be
+more than 180 days in the future.*, catching dates entered by mistake.
+
+The Beat task that launches due scheduled upgrades
+(``execute_scheduled_upgrades``) is registered in the deployment's own
+``CELERY_BEAT_SCHEDULE`` on a short cadence (60 seconds in production);
+see the docker-openwisp and ansible-openwisp2 recipes for the snippet.
```
nemesifier, line 164: Did you update the PRs in ansible-openwisp2 and docker-openwisp to include this yet?

docs/user/settings.rst
```diff
@@ -129,6 +129,37 @@ The Beat task that drives these reminders
...
+---------------------------------------------------
+
+============ =======================
+**type**:    ``int``
+**default**: ``15552000`` (180 days)
+============ =======================
+
+Furthest into the future, in seconds, that a mass upgrade may be
+scheduled. A time beyond it is rejected with *The scheduled time cannot be
+more than 180 days in the future.*, catching dates entered by mistake.
```
nemesifier: These paragraphs present the default values as if they were fixed validation messages. Deployments can override both settings, and the implementation derives the displayed minute/day bounds from the configured values, so operators reading an override configuration can be told to expect the wrong error text. This was a valid CodeRabbit suggestion and the current wording still does not state the distinction.

Please identify 10 minutes and 180 days as defaults in the prose and state that validation bounds and messages use the configured values.

Severity: P3

openwisp_firmware_upgrader/admin.py
```diff
@@ -138,33 +154,50 @@ class BatchUpgradeConfirmationForm(forms.ModelForm):
...
+        cleaned_data = super().clean()
+        # The browser posts a single UTC scheduled_at; parse it instead of the
+        # naive split-widget value.
+        scheduled_at = self.data.get("scheduled_at")
+        if scheduled_at:
+            parsed = parse_datetime(scheduled_at)
+            if parsed is None:
+                self.add_error("scheduled_at", _("Enter a valid date and time."))
+            else:
+                cleaned_data["scheduled_at"] = parsed
```
nemesifier, line 203: `parse_datetime()` accepts an ISO value without an offset and returns a naive datetime. This replaces the aware value produced by `SplitDateTimeField`; model validation then compares it with aware `timezone.now()` and raises `TypeError`, turning malformed admin input into HTTP 500 instead of a field error.

Please reject a naive hidden `scheduled_at` value or make it aware according to an explicit timezone rule, and add a form regression proving that a value without an offset never produces a server error.

Severity: P2

openwisp_firmware_upgrader/base/models.py
```diff
@@ -1021,6 +1255,160 @@ def notify_on_completion(cls, sender, instance, created, **kwargs):
...
+            UpgradeOperation = load_model("UpgradeOperation")
+            if not self.upgradeoperation_set.exists():
+                # No children yet: claim the row so the dispatched worker's CAS
+                # fails and it flashes nothing.
+                updated = self._meta.model.objects.filter(
+                    pk=self.pk, status="in-progress"
+                ).update(status="cancelled")
+                if updated:
+                    self.status = "cancelled"
+                    self.save(update_fields=["status"])
```
nemesifier: Cancellation is not atomic with worker startup. A worker can claim the `in-progress` row and start creating children in its transaction while this query still sees no committed children. Cancellation then writes `cancelled`, but the worker can overwrite that state at line 943 or commit and dispatch its new children afterward. The endpoint has reported success at that point, but devices can still be flashed. The current regression test cancels only before the worker enters `upgrade()`, so it misses this window.

Please serialize cancellation with the worker claim and child creation, and add a concurrency test that pauses the worker after its claim, cancels the batch, resumes the worker, and verifies that no operation is dispatched after cancellation succeeds.

Severity: P1

openwisp_firmware_upgrader/base/models.py
```diff
@@ -1021,6 +1255,160 @@ def notify_on_completion(cls, sender, instance, created, **kwargs):
...
+                continue
+            if batch.status != "scheduled":
+                continue
+            result = cls.dry_run(
+                build=batch.build, group=batch.group, location=batch.location
+            )
+            eligible = result["device_firmwares"].exists() or (
+                batch.firmwareless and result["devices"].exists()
+            )
+            conflict = batch._find_conflict()
```
nemesifier: The launch decision uses a stale batch snapshot while `reschedule()` locks only its own update. For example:

1. The scanner loads a due batch with `firmwareless=True` and validates it.
2. The operator changes `firmwareless` to `False` without moving the already-due time.
3. The scanner's status/time CAS still succeeds and dispatches the stale `True` argument.

The same race with `group` or `location` can validate the old population, execute the new empty population, and leave the batch `in-progress` without children.

Please lock and reload the batch while re-validating and claiming it, using the same row lock as `reschedule()`, and dispatch only values read inside that protected section. Add a regression for an editable field changing between scan and claim.

Severity: P1

openwisp_firmware_upgrader/base/models.py
```diff
@@ -679,24 +705,94 @@ def clean(self):
                     )
                 }
             )
-        self._validate_is_persistent_immutable()
+        stored = None if self._state.adding else self._fetch_stored_row()
+        self._validate_schedule(stored)
+        self._validate_is_persistent_immutable(stored)
+        self._validate_scheduled_editability(stored)
+        self._validate_no_conflict()
```
nemesifier, line 717: Conflict validation and persistence are separate operations, so concurrent requests can create overlapping active batches:

1. Both requests run `_find_conflict()` before either row exists.
2. Both see no conflict.
3. Both save and dispatch upgrades for the same population.

The sequential conflict tests cannot detect this, and a later per-device check cannot guarantee that duplicate flashing is prevented.

Please serialize every conflict-sensitive creation and reschedule by locking a stable shared row, such as the firmware category, around validation and save. Add a transaction test that synchronizes two requests before either save and proves only one active batch is created.

Severity: P1

openwisp_firmware_upgrader/base/models.py
```diff
@@ -1021,6 +1255,160 @@ def notify_on_completion(cls, sender, instance, created, **kwargs):
...
+                        batch.save(update_fields=["status"])
+                        batch._scheduled_validation_failed(reason=conflict)
+                continue
+            claimed = cls.objects.filter(
+                pk=batch_id, status="scheduled", scheduled_at__lte=now
+            ).update(status="in-progress")
+            if not claimed:
+                continue
+            try:
+                batch_upgrade_operation.delay(batch_id, batch.firmwareless)
```
nemesifier: The durable database state is changed to `in-progress` before broker publication. If this method runs in an outer transaction, the worker can see the old `scheduled` state and return before the claim commits. A process exit after the claim but before `delay()` also leaves the batch permanently `in-progress`, and a retry of the Beat task skips it. Wrapping `delay()` in `on_commit()` fixes the visibility case but not the crash gap.

Please keep the durable state retryable until publication, or preferably let an idempotent worker perform the `scheduled` to `in-progress` CAS and runtime validation. Add coverage for transaction visibility and interrupted publication so a due batch cannot be orphaned without children.

Severity: P1

openwisp_firmware_upgrader/base/models.py
```diff
@@ -706,9 +802,145 @@ def _validate_is_persistent_immutable(self):
...
+                    "when": when,
+                }
+        UpgradeOperation = load_model("UpgradeOperation")
+        device_ids = self.build._find_related_device_firmwares(
+            group=self.group, location=self.location
+        ).values_list("device_id", flat=True)
+        clashing = UpgradeOperation.objects.filter(
+            status__in=UpgradeOperation.CANCELLABLE_STATUS,
+            device_id__in=device_ids,
+        )
```
nemesifier, line 932: The per-device conflict query only considers devices returned by `_find_related_device_firmwares()`. A target included through `firmwareless=True` is absent from `device_ids`, so an active operation is missed if its `DeviceFirmware` row was removed while the operation remained active. The new batch can then create another operation for that device, contrary to the documented conflict guarantee.

Please include the firmwareless target queryset when building the device IDs whenever `self.firmwareless` is enabled, and add a regression covering an active operation on a currently firmwareless target.

Severity: P2

openwisp_firmware_upgrader/base/models.py
```diff
@@ -642,6 +650,24 @@ class AbstractBatchUpgradeOperation(
             "this batch, null if no reminder has been sent yet"
         ),
     )
+    scheduled_at = models.DateTimeField(
+        null=True,
+        blank=True,
+        db_index=True,
```
nemesifier: A standalone `scheduled_at` index does not match the minute-by-minute scanner query. Over time, `scheduled_at__lte=now` includes every completed historical scheduled batch, so the database must scan those index entries and filter by `status` on every run. The PR description and issue checklist originally called for `(status, scheduled_at)`, which supports the equality plus range predicate directly.

Please use a composite `(status, scheduled_at)` index, or a partial due-scan index if all supported databases can represent it, and assert the intended index in the migration test.

Severity: P3

openwisp_firmware_upgrader/settings.py
```diff
@@ -48,6 +48,18 @@
...
+    settings, "OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MIN_DELAY", 600
+)
+SCHEDULE_MAX_HORIZON = getattr(
+    settings, "OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MAX_HORIZON", 15552000
+)
+if SCHEDULE_MIN_DELAY >= SCHEDULE_MAX_HORIZON:
+    raise ImproperlyConfigured(
+        "OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MIN_DELAY must be smaller than "
+        "OPENWISP_FIRMWARE_UPGRADER_SCHEDULE_MAX_HORIZON"
+    )
```
nemesifier, line 69: The settings check validates only their ordering. A negative minimum such as `SCHEDULE_MIN_DELAY=-3600` passes startup and allows a timestamp 30 minutes in the past to pass `_validate_schedule()`; Beat then launches it immediately. Non-positive horizons can likewise leave no usable scheduling range.

Please require a non-negative minimum and a positive maximum in addition to `min < max`, with settings tests for invalid boundary values.

Severity: P2

openwisp_firmware_upgrader/static/firmware-upgrader/js/batch-actions.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+    $("#ow-batch-cancel-modal").removeClass("ow-hide");
+    $("#ow-batch-cancel-modal .ow-batch-cancel-confirm").trigger("focus");
+  }
+
+  function createBatchCancelModal() {
+    const modalHtml = `
+      <div id="ow-batch-cancel-modal" class="ow-overlay ow-overlay-notification ow-overlay-inner ow-hide">
+        <div class="ow-dialog-notification ow-cancel-confirmation-dialog">
+          <span class="ow-dialog-close ow-dialog-close-x">&times;</span>
```
nemesifier: The new confirmation overlay is not exposed as a modal dialog. It has no `role="dialog"`, `aria-modal`, or accessible label; the X close control is a non-focusable `span`; and focus is not restored when the overlay closes. Keyboard and screen-reader users cannot reliably discover or leave this destructive confirmation.

Please use a real button for the close control, add dialog semantics tied to the heading, trap focus while open, and restore focus to `#batch-cancel-btn` on close.

Severity: P3

openwisp_firmware_upgrader/static/firmware-upgrader/js/upgrade-selected-confirmation.js
```diff
@@ -13,4 +13,35 @@ django.jQuery(function ($) {
...
+  const scheduleRow = $("#schedule-row");
+  if (scheduleRow.length) {
+    const form = scheduleRow.closest("form");
+    const utcInput = $('<input type="hidden" name="scheduled_at">');
+    form.append(utcInput);
+    const dateInput = scheduleRow.find('input[name="scheduled_at_0"]');
+    const timeInput = scheduleRow.find('input[name="scheduled_at_1"]');
+    form.on("submit", function () {
+      const scheduled = new Date(dateInput.val() + "T" + timeInput.val());
+      utcInput.val(isNaN(scheduled.getTime()) ? "" : scheduled.toISOString());
```
nemesifier, line 29: Django's `AdminSplitDateTime` shortcuts deliberately fill `Now`, `Today`, and `Tomorrow` in the server timezone using `DateTimeShortcuts.now()`. This handler then interprets those values as browser-local time. When browser and server timezones differ, a shortcut-selected schedule is shifted by the offset and can even land on the wrong date. The reschedule handler has the same behavior.

Please replace or adapt the server-aware shortcuts so every displayed value really is browser-local before conversion, and add browser coverage with different server and browser timezones.

Severity: P2

openwisp_firmware_upgrader/templates/admin/firmware_upgrader/batch_upgrade_operation_change_form.html
```diff
@@ -12,6 +12,91 @@
...
+  <div id="batch-reschedule-form" class="batch-reschedule-form ow-hide"
+       role="region" aria-label="{% trans 'Edit schedule' %}"
+       data-server-tz="{{ server_timezone }}"
+       data-scheduled-utc="{{ batch.scheduled_at|date:'c' }}">
+
+    <div class="form-row">
+      <div>
+        <div class="flex-container">
+          <label class="required">{% trans "Scheduled time" %}:</label>
+          {{ reschedule_form.scheduled_at }}
```
nemesifier: The `Scheduled time` label is not associated with either split input. Unlike the normal Django form row, this manually rendered label has no `for` attribute, and `AdminSplitDateTime` does not add labels for `scheduled_at_0` and `scheduled_at_1`, so assistive technology receives two unnamed controls.

Please associate the visible label with the date input and provide a distinct accessible name for the time input, then cover those names in the rendered admin test.

Severity: P3

openwisp_firmware_upgrader/tests/test_selenium.py
```diff
@@ -560,6 +560,84 @@ def test_mass_upgrade_confirmation_page_widgets(self):
...
+            submit = self.find_element(
+                by=By.CSS_SELECTOR, value='input[name="upgrade_all"]'
+            )
+            self.web_driver.execute_script("arguments[0].click();", submit)
+            WebDriverWait(self.web_driver, 5).until(
+                EC.url_contains("batchupgradeoperation")
+            )
+            batch = BatchUpgradeOperation.objects.get(build=build2)
+            self.assertEqual(batch.status, "scheduled")
+            self.assertIsNotNone(batch.scheduled_at)
```
nemesifier: This assertion does not cover the browser-to-UTC behavior that makes the Selenium test valuable. Removing the conversion or shifting the instant by several hours would still pass. A direct comparison with `scheduled` is also unsafe unless the browser timezone is controlled, because Django and Chrome may use different zones.

Please set a known browser timezone that differs from `TIME_ZONE`, or derive the expected UTC instant from the browser, and assert the exact stored value. This endorses the remaining valid part of the existing bot suggestion.

Severity: P3

nemesifier (changes requested):
Good progress, see my comments below

Eeshu-Yadav:
@coderabbitai full review, focus on security, correctness, race conditions, efficiency

Eeshu-Yadav:
@coderabbitai full review, focus on security, correctness, race conditions, efficiency

nemesifier (approved):

nemesifier:
@coderabbitai resolve

